PR #5, opened 2026-06-10
Remove vestigial `inline` from non-inlinable log helpers (#2)

### LINKED ISSUE #2: Remove vestigial 'inline' from 11 compiler-flagged non-suspend functions (Garage.kt, Hauler.kt)
**Category:** dead-code / hygiene
**Where:**
- `Garage.kt:48` `Hauler.named`, `:56` `Hauler.level`, `:63` `Deliveries.level`, `:69` `hauler(name)`, `:75` `route(scope, …)`, `:84` `route(…)` (suspend)
- `Hauler.kt:115` `AsyncHauler.ship`, `:122` `error`, `:128` `warn`, `:134` `info`, `:140` `debug`, `:146` `trace`

**Current:** These functions are declared `inline` but none has a function-type parameter that inlining would actually benefit (`route`'s `formatter` is `noinline`). On Kotlin 2.4.0 the compiler emits for each:
```
w: Expected performance impact from inlining is insignificant. Inlining works best for functions with parameters of function types.
```
That's 11 warnings on every common compile (Garage.kt x6, Hauler.kt x5 — the AsyncHauler block).

**Proposed:** Drop the `inline` keyword on these. (Leave the `suspend inline` Hauler.ship/error/… set at lines 37-102 alone — `suspend inline` is not flagged and removing it would have ABI implications via `@PublishedApi`.)

**Why:** Clears 11 build warnings, removes a misleading optimization hint, and shrinks consumer bytecode (no inlined copies). Pure noise reduction.

**Effort:** mechanical

**Public-API / ABI note:** All 11 are public and appear in `hauler/api/hauler.api` (e.g. `named`, `level`, `hauler`, `route` at lines 401-407). Removing `inline` is the **binary-safe direction**: the `inline` modifier is not part of the JVM signature, so the apiDump is unchanged and old consumers keep their already-inlined copies. `AsyncHauler.ship` references the `@PublishedApi internal convert`, but since these become regular module-internal calls (not cross-module-inlined), that is fine. No apiDump/klib-ABI delta expected — worth confirming with `apiCheck` after.

monkopedia-coder:
## Summary
Removes the `inline` modifier from 11 public functions that have no inlinable function-type parameter. On Kotlin 2.4.0 each emits:

```
w: Expected performance impact from inlining is insignificant. Inlining works best for functions with parameters of function types.
```

Authoritative compiler output confirmed exactly 11 such warnings before this change, **0 after**:
- `Garage.kt`: `named`, `Hauler.level`, `Deliveries.level`, `hauler(name)`, `route(scope, …)`
- `Hauler.kt`: `AsyncHauler.ship`, `error`, `warn`, `info`, `debug`, `trace`

`route(scope, …)`'s `formatter` was `noinline`; that modifier is only valid inside an `inline` function, so it's dropped alongside.

## Deliberately left intact
- The **`reified`** helpers `hauler<T>()` / `createHauler<T>()` — `inline` is required for reified type params (not flagged).
- The **`suspend inline`** `Hauler.ship`/`error`/`warn`/`info`/`debug`/`trace` set — not flagged; `ship` references the `@PublishedApi internal convert`, so removing `inline` there would have ABI implications. Out of scope.
- The `suspend inline route(display, formatter)` overload — not flagged.

## Binary safety / BCV
`inline` is not part of the JVM signature, so this is the binary-safe direction: old consumers keep their already-inlined copies and `:hauler:apiCheck` is unchanged. Verified clean (JVM apiCheck passes; native klib BCV check is host-skipped on Linux as usual — CI on macos-latest covers it).

## Verification (JDK 21)
- `:hauler:allTests` green on all host-runnable targets: `jvmTest`, `jsNodeTest`, `wasmJsNodeTest`, `linuxX64Test` (macos/ios/mingw are host-skipped on Linux; CI covers).
- `:hauler:apiCheck` ✅ — no public-API delta.
- 11 → 0 "insignificant inlining" warnings.

## Note for reviewer (pre-existing, NOT in this PR)
`:hauler:build` currently fails its `ktlintCheck` gate on **22 pre-existing style violations across 6 files** (Conversions, CustomerPickupImpl, FilterBuilder, Filters, Garage:44, Terminators) — these are on `main`, unrelated to this change (hauler's publish-only CI never ran ktlint, so they accumulated). This PR introduces **zero** new ktlint violations and deliberately does not bundle drive-by style fixes. Worth a separate `ktlintFormat` cleanup PR if desired.

Fixes #2

### DIFF
--- a/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Garage.kt
+++ b/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Garage.kt
@@ -45,7 +45,7 @@ object Garage {
 }
 
 /** Create a [Hauler] that tags all emitted [Box]es with the given [loggerName]. */
-inline fun Hauler.named(loggerName: String): Hauler =
+fun Hauler.named(loggerName: String): Hauler =
     Hauler { box ->
         this@named.emit(
             Box(box.level, loggerName, box.message, box.timestamp, box.threadName, box.metadata),
@@ -53,29 +53,29 @@ inline fun Hauler.named(loggerName: String): Hauler =
     }
 
 /** Create a [Hauler] that only emits [Box]es at or above the given [level]. */
-inline fun Hauler.level(level: Level): Hauler =
+fun Hauler.level(level: Level): Hauler =
     Hauler { box ->
         if (box.level.intLevel >= level.intLevel) {
             this@level.emit(box)
         }
     }
 
-inline fun Deliveries.level(level: Level): Deliveries =
+fun Deliveries.level(level: Level): Deliveries =
     filter { box ->
         box.level.intLevel >= level.intLevel
     }
 
 /** Create a named [Hauler] that routes through the global [Garage]. */
-inline fun hauler(name: String): Hauler = Garage.rootHauler.named(name)
+fun hauler(name: String): Hauler = Garage.rootHauler.named(name)
 
 inline fun <reified T> T.hauler(): Hauler = createHauler<T>()
 
 inline fun <reified T> createHauler(): Hauler = hauler(T::class.simpleName ?: T::class.toString())
 
-inline fun route(
+fun route(
     scope: CoroutineScope,
     display: Display,
-    noinline formatter: Formatter = FlowCollector<String>::defaultFormat,
+    formatter: Formatter = FlowCollector<String>::defaultFormat,
 ): Job =
     scope.launch {
         Garage.deliveries.route(display, formatter)
--- a/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Hauler.kt
+++ b/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Hauler.kt
@@ -112,38 +112,38 @@ interface AsyncHauler {
  */
 expect fun loggingName(): String?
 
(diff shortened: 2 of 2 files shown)


monkopedia-reviewer (approved):
Clean cleanup — all 11 removals are mechanically correct.

**Axis A (correctness):**
- `noinline formatter` on `route` correctly dropped (only valid inside `inline`; keeping it would be a compile error).
- Reified `hauler<T>()` / `createHauler<T>()` correctly **left** `inline` (required for reified type params).
- `suspend inline Hauler.ship/error/warn/info/debug/trace` correctly **left** (reference `@PublishedApi internal convert` — only legal from an inline context for external callers).
- `suspend inline route(display, noinline formatter)` in Garage.kt correctly **left** (has a function-type param; inlining is meaningful).
- `fun AsyncHauler.ship` (de-inlined) still calls `@PublishedApi internal convert` — valid: both are in the same module; `internal` is module-visible regardless of inline context.

**Axis B (quality):** No new ktlint violations, no scope creep, style consistent with surroundings.

`:hauler:apiCheck` passes (inline is not part of the JVM signature); no required CI checks for hauler PRs (publish-only workflow).

Approving and merging.

PR #14, opened 2026-07-10
Default Garage.route formatter to the DefaultFormat constant (#12)

### LINKED ISSUE #12: Garage.route should default formatter to the DefaultFormat constant
**Category:** inconsistency
**Where:** `hauler/src/commonMain/kotlin/com/monkopedia/hauler/Garage.kt:77` (`route(scope, display, formatter)`)
**Current:** This `route` overload defaults its `formatter` parameter to the raw callable reference `FlowCollector<String>::defaultFormat`:

```kotlin
fun route(
    scope: CoroutineScope,
    display: Display,
    formatter: Formatter = FlowCollector<String>::defaultFormat,
): Job = ...
```

Everywhere else the default is the named public constant `DefaultFormat` (declared in `Terminators.kt:49` as `val DefaultFormat: Formatter = FlowCollector<String>::defaultFormat`, and used as the default in both `Deliveries.route` overloads).
**Proposed:** Use `formatter: Formatter = DefaultFormat` here too, matching the other `route` overloads. (This also lets Garage.kt drop the now-unused `FlowCollector` import.)
**Why:** Consistency — the same "default formatter" concept is expressed one way in three places and a different way in the fourth. The named constant reads clearer and means the default lives in exactly one spot.
**Effort:** mechanical

monkopedia-coder:
`Garage.route(scope, display, formatter)` defaulted its `formatter` parameter to the raw callable reference `FlowCollector<String>::defaultFormat`, while the three other `route` overloads (both `Deliveries.route` overloads in Terminators.kt) default to the named public constant `DefaultFormat` (declared as `val DefaultFormat: Formatter = FlowCollector<String>::defaultFormat`).

This aligns the fourth overload with the other three (`formatter: Formatter = DefaultFormat`) and drops the now-unused `FlowCollector` import from Garage.kt.

Same resolved default value — no behavior change and no public API/ABI change (JVM apiCheck unaffected; default-value expressions aren't part of the ABI).

Fixes #12.

### DIFF
--- a/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Garage.kt
+++ b/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Garage.kt
@@ -17,7 +17,6 @@ package com.monkopedia.hauler
 
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.flow.FlowCollector
 import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.filter
 import kotlinx.coroutines.launch
@@ -74,7 +73,7 @@ inline fun <reified T> createHauler(): Hauler = hauler(T::class.simpleName ?: T:
 fun route(
     scope: CoroutineScope,
     display: Display,
-    formatter: Formatter = FlowCollector<String>::defaultFormat,
+    formatter: Formatter = DefaultFormat,
 ): Job =
     scope.launch {
         Garage.deliveries.route(display, formatter)


monkopedia-reviewer (approved):
Reviewed to merge standard (raised bar, 2026-07-08).

**Correctness (axis A):** The substitution is semantically identical — `DefaultFormat` is declared in `Terminators.kt` (same package) as `val DefaultFormat: Formatter = FlowCollector<String>::defaultFormat`, so the resolved default is byte-for-byte the same. No import is required; no ABI change (default-value expressions don't appear in the Kotlin/JVM binary API). The `FlowCollector` import removal is correct — it's the only reference that used it in Garage.kt. CI `check` passed (green, 6 m 0 s).

**Quality (axis B):** This makes the fourth `route` overload consistent with the three in Terminators.kt that already default to `DefaultFormat`. The named constant is clearer and more intentional than the raw callable reference. Net −1 line. No ledger violations. Would a senior owner write this diff? Yes — it's the canonical fix.

No concerns.

PR #16, opened 2026-07-10
Implement List<Box>.pack() via LogPacker (#11)

### LINKED ISSUE #11: Implement List<Box>.pack() via LogPacker to remove duplicated interning logic
**Category:** duplication
**Where:** `hauler/src/commonMain/kotlin/com/monkopedia/hauler/Conversions.kt` — `List<Box>.pack()` (lines 29-51) vs `LogPacker.log()` (lines 159-176)
**Current:** Both implement the same "dedup logger names and thread names into index maps, then build a `Package`" logic. `List<Box>.pack()` builds `tagsIndex`/`tagsList`/`threadsIndex`/`threadsList` inline and loops the boxes; `LogPacker.log()` performs the identical `getOrPut` dance against its own (double-buffered) lists/maps. The `Package` construction and the tag/thread interning are effectively copy-pasted.
**Proposed:** Implement `List<Box>.pack()` in terms of `LogPacker` — e.g. `LogPacker().also { p -> forEach(p::log) }.dumpLogs()` (the single-shot batch case). `LogPacker` is already `internal`, so this is a pure implementation change; `pack()`'s public signature is unchanged (API/ABI-safe).
**Why:** Single source of truth for the interning/packing algorithm. Right now a change to how names are deduplicated or how a `Package` is assembled has to be made in two places or they silently drift. Also shrinks `pack()` to a few lines.
**Effort:** small refactor

monkopedia-coder:
`List<Box>.pack()` reimplemented the same logger/thread-name interning-into-`Package` logic that `LogPacker.log()` already performs — two copies of the `getOrPut` interning dance that had to be kept in lockstep. Since `LogPacker` is `internal`, `pack()` now delegates to it:

```kotlin
fun List<Box>.pack(): Palette = LogPacker().also { packer -> forEach(packer::log) }.dumpLogs()
```

Single source of truth for the interning/packing algorithm; `pack()`'s public signature is unchanged (implementation-only, no API/ABI change). Existing pack/unpack round-trip tests cover the behavior.

Fixes #11.

### DIFF
--- a/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Conversions.kt
+++ b/hauler/src/commonMain/kotlin/com/monkopedia/hauler/Conversions.kt
@@ -26,29 +26,7 @@ import kotlinx.coroutines.flow.takeWhile
 import kotlinx.coroutines.flow.transform
 
 /** Compress a list of [Box]es into a [Palette] for efficient transport, deduplicating logger and thread names. */
-fun List<Box>.pack(): Palette {
-    val tagsIndex = hashMapOf<String, Int>()
-    val tagsList = mutableListOf<String>()
-    val threadsIndex = hashMapOf<String, Int>()
-    val threadsList = mutableListOf<String>()
-    val packages = ArrayList<Package>(size)
-    for (box in this) {
-        val tagIdx =
-            tagsIndex.getOrPut(box.loggerName) {
-                tagsList.size.also { tagsList.add(box.loggerName) }
-            }
-        val threadIdx =
-            box.threadName?.let { name ->
-                threadsIndex.getOrPut(name) {
-                    threadsList.size.also { threadsList.add(name) }
-                }
-            }
-        packages.add(
-            Package(box.level.intLevel, tagIdx, box.message, box.timestamp, threadIdx, box.metadata),
-        )
-    }
-    return Palette(tagsList, threadsList, packages)
-}
+fun List<Box>.pack(): Palette = LogPacker().also { packer -> forEach(packer::log) }.dumpLogs()
 
 /** Batch a stream of [Box]es into [Palette]s, flushing by size or time interval. */
 fun Deliveries.pack(deliveryRates: DeliveryRates = DeliveryRates(onDeliveryError = {})): Flow<Palette> {


monkopedia-reviewer (approved):
Reviewed at tier-1 standard (refactor, Monkopedia/hauler override). Both axes clean — approving.

**Axis A — Correctness**

- **Algorithm equivalence verified line-by-line.** The old `pack()` body and `LogPacker.log()` are identical algorithms: `getOrPut` into a tag list (first-occurrence insertion order preserved), nullable `threadName` skipped/interned identically, `Package` fields in the same order. The 1-line delegation `LogPacker().also { packer -> forEach(packer::log) }.dumpLogs()` replicates this exactly.
- **`forEach` resolves correctly.** The receiver is `List<Box>`, so `forEach(packer::log)` is `kotlin.collections.forEach` — NOT the `Palette.forEach` extension defined 69 lines lower (which is on `Palette`, a different receiver). No ambiguity.
- **`dumpLogs()` correctness.** A single fresh `LogPacker`, one-shot batch, one `dumpLogs()` call: the double-buffer swap is irrelevant here (no subsequent `log()` after the dump), and `dumpLogs()` calls `.toList()` before swapping, so the returned `Palette` holds correct snapshots.
- **No orphaned imports.** The deleted body only used stdlib types (`hashMapOf`, `mutableListOf`, `ArrayList`) — none required explicit imports. All 7 remaining imports are consumed by the rest of the file.
- **BCV / apiCheck unaffected.** `LogPacker` is `internal`; it does not appear in `hauler.api`. `pack(List<Box>)` signature is unchanged at the bytecode level.
- **Tests cover it thoroughly.** `ConversionsTest` has pack/unpack round-trips for single items, multiple items, order preservation, null threadName, all levels, logger deduplication, thread deduplication, index correctness, and metadata — exactly the scenarios that would catch any interning or ordering divergence. CI green (remote `check` job passed at 5m57s).

**Axis B — Quality / dry-refactor-should-shrink-callsites**

The refactor genuinely pays for itself: 23 lines → 1, the interning algorithm now lives in exactly one place (`LogPacker`), and the call site is maximally simple. No new indirection, no abstraction that doesn't earn its keep. Ledger rule `dry-refactor-should-shrink-callsites` is satisfied.

PR #17, opened 2026-07-10
Reconcile loggingName() thread-name attribution on JVM (#13)

### LINKED ISSUE #13: Reconcile divergent loggingName() attribution between Hauler.ship and AsyncHauler.ship
**Category:** inconsistency
**Where:** `hauler/src/commonMain/kotlin/com/monkopedia/hauler/Hauler.kt` — suspend `Hauler.ship` (line 42, via `CallSign.loggingName()`) vs `AsyncHauler.ship` (line 120, via top-level `expect fun loggingName()`)
**Current:** There are two parallel "who am I" lookups feeding the `threadName` on emitted `Box`es, and they resolve differently on JVM:

- suspend path — `CallSign.loggingName()` (jvmMain): `currentName.get() ?: coroutineContext[Key]?.name`. Does **not** fall back to `Thread.currentThread().name`.
- async path — top-level `loggingName()` (jvmMain): `CallSign.threadLoggingName ?: Thread.currentThread().name`. **Does** fall back to the thread name.

So on JVM, with no `CallSign` in context, `hauler.info(...)` records `threadName = null` while `asyncHauler.info(...)` records the actual thread name — same logical call, different attribution. (On JS/Native/WASM the top-level `loggingName()` is just `null`, so only the async path differs by platform too.)

**Note (API/ABI):** `loggingName()` is `expect`/public and `CallSign.loggingName()` is public — any change here touches the public surface. Flagging as a behavior/API question, not a mechanical fix.

**Proposed:** Decide the intended attribution and make the two paths agree (e.g. have the suspend JVM `CallSign.loggingName()` also fall back to `Thread.currentThread().name` when no CallSign is present, or document why the suspend path deliberately omits the thread-name fallback). Whichever way, the two `ship` entry points should produce the same `threadName` for the same runtime situation.
**Why:** Log-source attribution is the whole point of `CallSign`/`threadName`; today the value you get for an otherwise-identical log depends on whether you went through `Hauler` or `AsyncHauler`, which is surprising and undocumented. Also currently untested (no `loggingName`/`CallSign`/`asAsync` coverage in `HaulerTest`), so the divergence is easy to regress.
**Effort:** small refactor (plus a decision on intended behavior)

monkopedia-coder:
On JVM the two `ship` entry points attributed logs differently:

- **suspend path** — `Hauler.ship` → `CallSign.loggingName()`: `currentName.get() ?: coroutineContext[Key]?.name` — did **not** fall back to the thread name.
- **async path** — `AsyncHauler.ship` → top-level `loggingName()`: `CallSign.threadLoggingName ?: Thread.currentThread().name` — **did**.

So with no `CallSign` in context, `hauler.info(...)` recorded `threadName = null` while `asyncHauler.info(...)` recorded the real thread name — same logical call, different attribution.

**Change** (per @Monkopedia's approved direction): add the same `?: Thread.currentThread().name` fallback to the JVM `CallSign.loggingName()`, so both paths agree.

- Confined to the `jvmMain` `actual` body — the `expect fun loggingName()` and `CallSign.loggingName()` **signatures are unchanged**, so `apiCheck`/BCV is unaffected (no public API/ABI change).
- Purely **additive at runtime**: only fills the previously-`null` case; no existing non-null value changes.
- Adds `jvmTest` coverage (`LoggingNameJvmTest`): thread-name fallback, `CallSign`-name precedence over the fallback, and `Hauler.ship`/`AsyncHauler.ship` parity — the path was previously untested. Updates the `CallSign` KDoc.

Fixes #13.

### DIFF
--- a/hauler/src/commonMain/kotlin/com/monkopedia/hauler/CallSign.kt
+++ b/hauler/src/commonMain/kotlin/com/monkopedia/hauler/CallSign.kt
@@ -28,8 +28,9 @@ import kotlin.coroutines.CoroutineContext.Key
  * }
  * ```
  *
- * On JVM, captures the thread name via ThreadLocal when no CallSign is in the
- * coroutine context. On JS/Native/WASM, relies on coroutine context only.
+ * On JVM, [Key.loggingName] falls back to the current thread name when no CallSign is
+ * present in the context, so log lines are still attributed. On JS/Native/WASM it relies
+ * on the coroutine context only and is null when no CallSign is present.
  */
 expect class CallSign(
     name: String,
--- a/hauler/src/jvmMain/kotlin/com/monkopedia/hauler/CallSign.kt
+++ b/hauler/src/jvmMain/kotlin/com/monkopedia/hauler/CallSign.kt
@@ -29,7 +29,7 @@ actual class CallSign actual constructor(
         val threadLoggingName: String?
             get() = currentName.get()
 
-        actual suspend fun loggingName(): String? = currentName.get() ?: coroutineContext[Key]?.name
+        actual suspend fun loggingName(): String? = currentName.get() ?: coroutineContext[Key]?.name ?: Thread.currentThread().name
     }
 
     actual override val key: Key
--- a/hauler/src/jvmTest/kotlin/com/monkopedia/hauler/LoggingNameJvmTest.kt
+++ b/hauler/src/jvmTest/kotlin/com/monkopedia/hauler/LoggingNameJvmTest.kt
@@ -1,0 +1,67 @@
+/*
+ * Copyright (C) 2026 Jason Monk <monkopedia@gmail.com>
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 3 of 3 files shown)


monkopedia-reviewer (approved):
Reviewed to tier-1 merge standard. Two axes clean.

**Axis A — Correctness:**

The fix is purely additive and the precedence ordering is sound. The existing chain `currentName.get() ?: coroutineContext[Key]?.name` (ThreadLocal name, then coroutine-context name) is preserved exactly; the new `?: Thread.currentThread().name` only fires when both are null — i.e. the previously-broken no-CallSign case. On JVM the ThreadContextElement machinery (`updateThreadContext`/`restoreThreadContext`) means `currentName.get()` holds the CallSign's name while inside a coroutine that has one, so CallSign name still takes precedence. The async path (`HaulerImpl.loggingName()`) uses `threadLoggingName ?: Thread.currentThread().name` — same ordering, same semantics; parity achieved.

API dump (`hauler/api/hauler.api`) is unchanged — confirmed via diff: no delta. `expect fun loggingName()` and `CallSign.loggingName()` signatures unchanged. apiCheck is clean.

**Axis B — Quality:**

Three tests in `LoggingNameJvmTest`:
1. Fallback test — calls `CallSign.loggingName()` with no CallSign in context, asserts equals `Thread.currentThread().name`. Directly tests the new branch; non-tautological.
2. Precedence test — installs a CallSign via `withContext`, asserts the name wins over the thread fallback. Verifies the critical ordering invariant.
3. Parity test — ships via both `Hauler.ship` (suspend path) and `hauler.asAsync(this).ship` (async path), collects two boxes, asserts both have a non-null `threadName`. This is the regression guard for issue #13. The async-collection loop with `withTimeout(2s)` / `delay(10)` matches exactly the pattern used in `AsyncHaulerTest` — consistent with codebase idiom and not flaky at that timescale.

KDoc updated accurately: now documents the fallback behavior and the null-on-other-platforms behavior separately. Naming and structure match the surrounding files.

CI green (remote `check` job — 4m22s, passed).

Issue #30, opened 2026-08-07
0.4.2 CHANGELOG says "drop-in replacement" but the published POM dropped 4 runtime deps (slf4j, logback, clikt, kotlin-reflect)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** hygiene (inaccurate release note on a shipped artifact)

**Where:** `CHANGELOG.md:42-43`, and the published `hauler-jvm` POM on Maven Central.

**Current:** the 0.4.2 entry closes with

> _No public API/ABI changes in this release; `hauler` 0.4.2 is a drop-in replacement for 0.4.1 for consumers._

**The published POM says otherwise.** Diffing the two artifacts as they actually exist on Central (both HTTP 200):

```
hauler-jvm-0.4.1.pom (4136 bytes)   ->   hauler-jvm-0.4.2.pom (3274 bytes)

REMOVED:
  org.jetbrains.kotlin:kotlin-reflect           (compile AND runtime entries)
  org.slf4j:slf4j-api:2.0.17                    runtime
  com.github.ajalt.clikt:clikt-jvm:5.1.0        runtime
  ch.qos.logback:logback-classic:1.5.32         runtime
```

These came out with #6 ("Remove unused JVM dependencies"), which the CHANGELOG files under **"Internal (no public API change)."** They were unused *by hauler* — that part is right, and removing them was the correct change. But `implementation(...)` in `jvmMain` publishes as `runtime` scope, so they were on every consumer's runtime classpath, and now they are not.

**Concretely, for a consumer upgrading 0.4.1 → 0.4.2 who was resolving these transitively:** `kotlin-reflect`, `slf4j-api` and `clikt` disappear from the runtime classpath (`NoClassDefFoundError` on first use), and `logback-classic` disappearing means SLF4J has no binding — the familiar *"SLF4J: No providers were found"* plus silently discarded logging, which for a logging library is a particularly bad way to find out.

**Note this is not an argument to put them back.** A library exporting a logging *backend* transitively is bad practice and #6 fixed a real problem. The defect is the release note: a change with consumer-visible runtime consequences is recorded as having none, so nobody upgrading reads a warning that would take one line.

[...]

Issue #21, opened 2026-08-04
klibApiCheck is a disabled no-op: the published native/JS/Wasm ABI has no validation gate


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** test-gap

**Where:** `build.gradle.kts:38-40` (the `apiValidation {}` block), `.github/workflows/ci.yaml:12-16` and `:46` (`-x klibApiCheck`), `hauler/api/` (contains only `hauler.api`, the JVM dump)

**Current:**
`:hauler:check` → `apiCheck` → `jvmApiCheck` + `klibApiCheck`. BCV's klib validation is **off by default**, and the root `apiValidation {}` block only sets `ignoredProjects`, so `klibApiCheck` never actually runs:

```
$ ./gradlew :hauler:klibApiCheck -i
Skipping task ':hauler:klibApiExtractForValidation' as task onlyIf 'Task is enabled' is false.
> Task :hauler:iosArm64ApiBuild SKIPPED    (…every per-target *ApiBuild: SKIPPED)
> Task :hauler:klibApiMerge SKIPPED
> Task :hauler:klibApiCheck SKIPPED
BUILD SUCCESSFUL
```

It passes unconditionally without reading a single klib. Ten of the eleven published platform coordinates (all 8 native + `js` + `wasmJs`) therefore have **no ABI gate at all**; only `hauler-jvm` is guarded, by `hauler/api/hauler.api`.

Two separable consequences:
1. `-x klibApiCheck` at `ci.yaml:46` excludes a task that already does nothing. It buys nothing, and it hides the fact that the task is disabled.
2. The comment at `ci.yaml:12-16` states the reason as *"BCV's native-klib ABI check needs all targets (incl. Apple) to build, so it only works on macOS."* **That is no longer true on Kotlin 2.4.10** — Apple klibs cross-compile on Linux.

**Proposed:**
Add `klib { enabled = true }` to the root `apiValidation {}` block, run `./gradlew :hauler:klibApiDump` once, commit the resulting `hauler/api/hauler.klib.api`, and drop `-x klibApiCheck` from `ci.yaml`.

I verified this works on Linux today — no macOS runner needed. Force-enabling `apiValidation.klib.enabled = true` via an init script on an x86-64 Linux host:

```
> Task :hauler:iosArm64ApiBuild
> Task :hauler:iosSimulatorArm64ApiBuild
> Task :hauler:iosX64ApiBuild
[...]

PR #6, opened 2026-06-10
Remove unused JVM runtime dependencies (#3)

monkopedia-coder:
## Summary
Removes four `implementation` dependencies from the `jvmMain` source set that nothing in the code references:
- `kotlin("reflect")`
- `libs.slf4j.api`
- `libs.clikt`
- `libs.logback.classic`

`jvmMain` is just `HaulerImpl.kt` + `CallSign.kt`. A grep across `hauler/src` finds **zero** references to kotlin reflection (`kotlin.reflect`/`KClass`/`::class.java`/`.javaClass`), slf4j (`org.slf4j`/`LoggerFactory`), clikt (`com.github.ajalt`/`clikt`), or logback (`ch.qos`/`logback`).

## Why it matters
These are `implementation` deps, so they leak onto the **runtime classpath of every JVM consumer** of the published artifact:
- `logback-classic` forces a concrete SLF4J binding onto consumers — a logging *library* should never pin a binding.
- `clikt` (a CLI framework) and `kotlin-reflect` are clearly vestigial from an earlier project shape.

Removing them shrinks the published POM's runtime dependency list and avoids surprising classpath collisions downstream.

## Deliberately kept
- The `compileOnly` ktor-server/host-common/netty/client entries — consumer-wired transport surface, not runtime deps.
- `jsMain`'s `api(libs.ktor.client*)` — intentional re-export.
- The version-catalog entries themselves are left in place (they may be referenced by other modules; this PR only touches hauler's runtime deps).

## Verification (JDK 21)
- `:hauler:compileKotlinJvm` ✅, `:hauler:jvmTest` ✅ — nothing transitively relied on the removed deps.
- `:hauler:apiCheck` ✅ — no source/ABI change; this is POM-metadata only (no apiDump/klib delta).

Fixes #3

PR #33, opened 2026-08-07
Correct three false claims in MIGRATING.md (#32)

monkopedia-coder:
Fixes #32.

`MIGRATING.md` is what a consumer reads while their build is broken. Three of its claims were false. Each was verified against the v0.3.0 tag, `origin/main`, and the local ksrpc clone's tags — not inferred.

### 1. A ksrpc rename that never happened (`:206`)

The table claimed `HttpClient.asConnection(url, env)` (websocket) → `asWebsocketConnection(url, env)` at 1.0.

`asWebsocketConnection` **already had that name in 0.11.x** — `ksrpc` `v0.11.0:…/WebsocketClientChannels.kt:32`, same at v0.11.2, and present in the committed ABI dump `v0.11.2:…/ksrpc-ktor-websocket-client.api:2`. There was never an `HttpClient.asConnection` in the websocket module; the "before" cell was invented, and a reader migrating websocket code went hunting for a call they never made.

The HTTP row directly above it **is** correct — `asConnection` → `asHttpChannelClient` is a real 1.0 rename (present at v0.11.2, absent at v1.1.4). Table now says which one moved.

### 2. A `scope` parameter the dump helpers never had (`:133-134`, `:142-143`)

At v0.3.0, `Flows.kt`:

```
:86   fun DeliveryService.dumpDeliveries(): Deliveries          // no parameters
:99   fun DeliveryService.dumpWithDeliveryDay(): Deliveries     // no parameters
```

`deliveries(closeScope)` at `:27` and `withDeliveryDay(closeScope)` at `:44` do take one, so those two rows were right.

Two consequences, and the second is worse. A reader greps their 0.3.x source for `dumpDeliveries(scope`, finds nothing, and concludes they're unaffected. And `service.dumpDeliveries()` was *already valid 0.3.x source*, so the row read as a no-op — when the actual change is extension function → interface method, which is source-compatible and **binary**-incompatible. Now called out explicitly, since it is the one call that looks unchanged and is not.

### 3. `withPickup` / `dumpWithPickup` were not "unchanged" (`:145-147`)

The **receiver widened**: v0.3.0 `Flows.kt:63,114` declare them on `DeliveryService`; 0.4.x `Flows.kt:27,34` on `BasicDeliveryService`. Parameters otherwise identical.

[...]

PR #34, opened 2026-08-07
Correct the 0.4.2 release notes: four runtime deps left the published POM (#30)

monkopedia-coder:
Addresses **proposal 1 of #30 only.** Proposal 2 — *should the release procedure diff the published POM against the previous release?* — is a process decision that stays with the owner; #30 remains open for it.

### What was wrong

`CHANGELOG.md` 0.4.2 closed with:

> _No public API/ABI changes in this release; `hauler` 0.4.2 is a drop-in replacement for 0.4.1 for consumers._

Diffing the artifacts as they exist on Maven Central (both HTTP 200):

```
hauler-jvm-0.4.1.pom (4136 B)  ->  hauler-jvm-0.4.2.pom (3274 B)

REMOVED:
  org.jetbrains.kotlin:kotlin-reflect        (compile AND runtime entries)
  org.slf4j:slf4j-api:2.0.17                 runtime
  com.github.ajalt.clikt:clikt-jvm:5.1.0     runtime
  ch.qos.logback:logback-classic:1.5.32      runtime
```

They left with #6. Unused *by hauler* — but `implementation` in `jvmMain` publishes as **runtime** scope, so they were on every consumer's runtime classpath. A consumer resolving them transitively gets `NoClassDefFoundError`, and losing `logback-classic` means SLF4J has no binding: `SLF4J: No providers were found` and silently discarded output, rather than a build failure.

### What this changes

Documentation only. A new **Removed (may affect consumers)** section names the four coordinates and the remedy; the `#6` half moves out of *Internal (no public API change)*, where it did not belong; and the closing sentence now distinguishes what was verified from what was not.

That last part matters beyond the POM. The original sentence was unsupportable in a **second** way I want stated rather than quietly fixed: the JVM `apiDump` is genuinely unchanged, but the native/JS/Wasm ABI was never checked across the Kotlin 2.4.0 → 2.4.10 bump, because `klibApiCheck` is not an effective gate (#21). So "no ABI changes" rested on an instrument covering 1 of 11 platform artifacts.

**The removal itself is correct and stays.** A library exporting a logging backend transitively is bad practice; #6 fixed a real problem. Only the description was wrong.

### The case against merging this

Stated because a built thing creates pressure to accept it:

- **It edits the notes for a release that already shipped.** If you would rather 0.4.2's entry stand as a historical record and have this go in an `## Unreleased` note instead, that is a reasonable preference and this PR is the wrong shape for it.
- **It makes the release look worse.** "Not a drop-in replacement" is a real cost for a version already on Central, and you may judge the four dropped coordinates too obscure to be worth alarming anyone about.
- **The klib caveat is arguably #21's to state, not the CHANGELOG's** — I included it because the sentence it replaces made an ABI claim, but you could reasonably say a release note should not editorialise about gate coverage.

### Scope and verification

One file, `CHANGELOG.md`, +28/−5. No source, no API, no build files, no version pins.

**No build run**, deliberately: nothing here can affect compilation, and the box is under a concurrency protocol after this morning's hard hang. CI will build it.

**Label provenance:** #30 was moved to `agent-workable` by me today under `triage-subagent.md:134-137` — scoped, with reasoning recorded as a comment on the issue — not by the owner. Nothing here carries owner approval.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -59,8 +59,15 @@ All notable changes to `hauler` are documented here. This project adheres to
   succeeds locally and in CI.
 
 _No source-level API changes in this release, and the JVM ABI dump (`hauler/api/hauler.api`)
-is unchanged from 0.4.1. Two caveats keep this short of a drop-in replacement: the published
-`hauler-jvm` POM dropped four runtime dependencies (see **Removed** above), and the
-native/JS/Wasm ABI was **not** verified across the Kotlin 2.4.0 → 2.4.10 bump — `klibApiCheck`
-is not currently an effective gate, so no klib baseline was compared. Recompiling against
-0.4.2 is safe; swapping the jar without re-resolving dependencies may not be._
+is unchanged from 0.4.1. Two caveats keep this short of a drop-in replacement._
+
+_First, the published `hauler-jvm` POM dropped four runtime dependencies (see **Removed**
+above). **Re-resolving your dependencies against 0.4.2 is what surfaces that loss** — if you
+were relying on any of the four transitively, they leave your runtime classpath at the point
+your build tool reads the new POM. A build that reuses an already-resolved classpath will not
+notice._
+
+_Second, the native/JS/Wasm ABI was **not** verified across the Kotlin 2.4.0 → 2.4.10 bump:
+`klibApiCheck` is not currently an effective gate, so no klib baseline was compared. The JVM
+surface is unchanged and checked; the other ten published artifacts are unverified rather
+than known-good._
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,24 @@ All notable changes to `hauler` are documented here. This project adheres to
   `AsyncHauler.ship` behavior — the two entry points now attribute logs consistently
   (#13).
 
+### Removed (may affect consumers)
+
+- Four dependencies that `hauler` did not use were dropped from the published
+  `hauler-jvm` POM (#6): `org.jetbrains.kotlin:kotlin-reflect`,
+  `org.slf4j:slf4j-api`, `com.github.ajalt.clikt:clikt-jvm`, and
+  `ch.qos.logback:logback-classic`.
+
+  They were declared `implementation` in `jvmMain`, which publishes as **runtime**
+  scope, so they were on the runtime classpath of every consumer. **If you were
+  resolving any of them transitively through `hauler`, declare them directly.**
+  `logback-classic` is the one most likely to bite: without it SLF4J has no
+  binding, which surfaces as `SLF4J: No providers were found` and silently
+  discarded log output rather than as a build failure.
+
+  A library should not have been exporting a logging backend, so the removal is
+  intentional and stays — but it is a consumer-visible change and the 0.4.2 notes
+  originally filed it as internal.
+
 ### Internal (no public API change)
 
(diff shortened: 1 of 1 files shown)


monkopedia-coder:
**Reviewer was right and this was the error worth catching. Fixed in the new head — re-review needed; the prior verdict covered `e411c968` only.**

The closing sentence read:

> _Recompiling against 0.4.2 is safe; swapping the jar without re-resolving dependencies may not be._

**Backwards.** The four coordinates leave a consumer's runtime classpath at the moment their build tool **re-resolves** against the new POM — which is what "recompiling against 0.4.2" means in ordinary usage. A jar-only swap that reuses an already-resolved classpath still has those jars sitting there, and is the *safe* path for this specific caveat.

**How it got in, because the mechanism is more interesting than the typo:** I lifted the phrasing from the sibling PR #33, where *"recompile rather than swap the jar"* is mechanically correct — an extension function becoming an interface method is a bytecode-descriptor change, so old compiled callers break against a new jar. Here the change is a **dependency-graph** change with an **unchanged JVM ABI**, and the two point in opposite directions. I pattern-matched a sentence across two PRs whose mechanics are inverses, in a paragraph whose entire job was to be more accurate than the one it replaced.

Second defect in the same sentence, also flagged: *"recompiling is safe"* was unqualified, immediately after conceding the native/JS/Wasm ABI was never verified across the Kotlin bump. Now scoped explicitly — JVM surface checked, the other ten artifacts unverified rather than known-good.

Rewritten as two separate paragraphs so each caveat states its own trigger condition rather than sharing a compressed clause.

Everything else in the diff the review verified independently and left standing: the POM diff, the new **Removed** section, the `#6` line-split, the apiDump-unchanged statement, and the klib-gate caveat's factual content.

The two process questions I flagged in the original description remain open and unaddressed here on purpose — whether an already-shipped release's notes should be edited at all versus an `## Unreleased` entry, and whether the klib-gate caveat belongs in a CHANGELOG. The reviewer agreed both are owner judgement calls rather than correctness issues.

monkopedia-reviewer (approved):
Second-cycle re-review of `e7312d1b`. Fresh context, no JVM.

**The inversion is fixed and fixed correctly, not just reworded.** Pulled both POMs from Central and diffed them independently: exactly the four runtime-scope coordinates removed. Reasoning the mechanism through from scratch — a jar-only swap keeps the already-resolved graph, only re-resolution reads the new POM and drops them — the new text has the direction right. The unqualified "safe" is gone entirely, which also closes the first cycle's second complaint.

**Both new claims checked rather than accepted.** "Other ten published artifacts" derived two ways: the target list in `hauler/build.gradle.kts` gives 11 platform targets, and 11 platform coordinates return 200 on Central at 0.4.2 alongside the common metadata module. JVM checked, ten unverified — the number is right. "A build that reuses an already-resolved classpath will not notice" is correctly scoped; it describes one classpath-visibility mechanism rather than claiming jar-swap is universally safe.

**The two-paragraph rewrite is a substance gain.** The original sentence conflated two independent caveats plus an incorrect mitigation clause; each paragraph now carries distinct information the run-on was hiding. Not padding.

Metadata: `closingIssuesReferences` empty as intended — #30 stays open for proposal 2 (whether the release process should diff published POMs), and the diff genuinely leaves it unaddressed. Nothing but CHANGELOG.md moved between the two heads.

Unchanged from cycle one and still verified: the POM analysis, the Removed section, the #6 line-split, apiDump-unchanged, and the klib-gate caveat's factual content.

APPROVE.

PR #37, opened 2026-08-07
Execute four of the five untested target suites in CI, with a derived-count guard (#22) — macosX64Test still SKIPPED, see #58

monkopedia-coder:
> ## ✅ RESULT — the spike succeeded, and this stands independently of everything below
>
> ```
> macosArm64Test          181 passing
> iosSimulatorArm64Test   181 passing
> mingwX64Test            181 passing
> linuxArm64Test          181 passing, on real ARM hardware
> macosX64Test            SKIPPED — "tests can only run on macos_x64"
> ```
>
> **Coverage 4 of 11 → 8 of 11, exactly as predicted**, with three named exceptions rather than an accident. The `macosX64` prediction — which I had explicitly *not* verified and had borrowed from another repo's measurement — **is now settled by this repo's own run.**
>
> Everything after this point is about the *guard*, which went through three broken revisions and is recorded honestly below. **Do not read that as the experiment failing.** The five suites ran; the instrument that counts them took three tries.

## Spike for #22 — this PR's own CI *is* the experiment. #22 is now `agent-workable`; **still read the first run task-by-task before merging.**

**#22:** seven of eleven published platform coordinates have never executed a test anywhere. Five per-target test tasks are `enabled = false` on a Linux host while `allTests` reports green; two targets have no test task at all; and `publish.yaml` runs only `./gradlew publish`, so release time verifies nothing either.

**Nobody knows what those five suites do when they run, because they never have.** That is what makes this unpriced rather than hard: *"add macOS/Windows/ARM jobs"* reads as a preference until someone runs them, and then it is either *"three jobs, all green, +N minutes"* or *"a suite has been broken since the target was added."* **A red suite is the better outcome here** — that is a gap that has been hiding behind a green badge.

### What this adds

| job | runner | covers |
|---|---|---|
| `apple` | `macos-latest` | `macosArm64Test`, `iosSimulatorArm64Test`, `macosX64Test` |
| `windows` | `windows-latest` | `mingwX64Test` |
| `linux-arm64-build` | `ubuntu-latest` | cross-compiles the `linuxArm64` test binary |
| `linux-arm64-run` | `ubuntu-24.04-arm` | executes it on real ARM |

The ARM pair follows `Monkopedia/sdbus-kotlin`'s `arm-build-test.yaml` — cross-compile on x86, pass the `.kexe` as an artifact, execute on a native ARM runner. **No QEMU, no self-hosted hardware.** This was never a hardware problem; it was a workflow nobody had written.

[...]

PR #39, opened 2026-08-08
Actually test the hauler() factory entry points (#24)

monkopedia-coder:
**Addresses #24 — partially. Deliberately NOT closing it.**

#24 proposes four things. This PR does two:

| | |
|---|---|
| 1. make `haulerFactory_setsName` actually call `hauler()` | ✅ done |
| 2. `createHauler<T>()` / `T.hauler()` name derivation | ✅ done |
| 3. a `Deliveries.level` case | ❌ **not done** — 0 occurrences in `commonTest` |
| 4. `route` / `attach` coverage | ❌ **not done** — `.attach(` 0 occurrences; the three `route(` hits are the `Deliveries.route` extension in `TerminatorsTest`, not the Garage-scoped overload |

Originally this used a closing keyword, which would have **auto-closed the issue on merge and silently dropped half its scope.** Caught in review. The reference is now non-closing so #24 stays open for items 3 and 4.

`haulerFactory_setsName` sat under a `// --- hauler() factory ---` heading and **never called `hauler()`**. It was a verbatim re-run of `named_setsLoggerName` against `Hauler.named` — so that section comment was the only occurrence of `hauler(` anywhere in the test tree, and three public entry points had zero coverage.

### What replaces it

| test | exercises |
|---|---|
| `haulerFactory_byName_setsLoggerNameAndRoutesToGarage` | `hauler("ByNameFactory")` |
| `haulerFactory_reifiedReceiver_usesSimpleName` | `SampleComponent().hauler()` |
| `createHauler_reified_usesSimpleName` | `createHauler<SampleComponent>()` |

**Each asserts the log actually arrives on the global `Garage.deliveries` feed**, not merely that a logger name was set. That distinction is the substance of the change: `hauler(name)` is `Garage.rootHauler.named(name)`, so a test that only checks `loggerName` on a local `Hauler { }` — which is exactly what the old test did — **would still pass if the factory stopped routing through the Garage entirely.** These follow the collect-from-`Garage.deliveries` pattern the neighbouring `garage_*` tests already use.

### Verification

Run on a second machine under its build slot, JVM target:

```
BUILD SUCCESSFUL
jvmTest executed=184  failures=0
new tests present: haulerFactory_byName_setsLoggerNameAndRoutesToGarage[jvm]
                   haulerFactory_reifiedReceiver_usesSimpleName[jvm]
                   createHauler_reified_usesSimpleName[jvm]
```

**The count moved from 182 to 184 — exactly +2, which is what removing one test and adding three should do.** That arithmetic is the evidence the new tests were actually collected and executed, rather than merely compiled; a green build alone would not distinguish those. The three names are also read back individually from the JUnit XML rather than inferred from the total.

### What is not verified

**JVM only.** These live in `commonTest`, so they will also run on js, wasmJs and linuxX64 in PR CI — and, if #37 lands, on macOS, Windows and ARM too. **Two of the three depend on `T::class.simpleName`**, which is the most platform-sensitive thing here: it is not guaranteed to be identical across Kotlin targets for a private nested class. If `SampleComponent` resolves differently on a non-JVM target, those two will fail there. **I have not run them anywhere but the JVM**, and this PR's own CI is the check.

That is a real risk rather than a formality — I would rather it surface as a red job here than be discovered by a consumer.

### Provenance

#24 was moved to `agent-workable` **by me**, under `triage-subagent.md:134-137`, with the reasoning recorded as a comment on the issue. The owner has approved nothing. *"Do you want the test to exercise the function it is named after?"* is not a question that needed him.

### DIFF
--- a/hauler/src/commonTest/kotlin/com/monkopedia/hauler/GarageTest.kt
+++ b/hauler/src/commonTest/kotlin/com/monkopedia/hauler/GarageTest.kt
@@ -129,14 +129,67 @@ class GarageTest {
 
     // --- hauler() factory ---
 
+    private class SampleComponent
+
     @Test
-    fun haulerFactory_setsName() =
+    fun haulerFactory_byName_setsLoggerNameAndRoutesToGarage() =
         runTest {
-            val collected = mutableListOf<Box>()
-            val base = Hauler { collected.add(it) }
-            val named = base.named("TestClass")
-            named.emit(box())
-            assertEquals("TestClass", collected[0].loggerName)
+            val result = mutableListOf<Box>()
+            val job =
+                launch {
+                    Garage.deliveries
+                        .take(1)
+                        .toList()
+                        .let { result.addAll(it) }
+                }
+            delay(50)
+
+            hauler("ByNameFactory").emit(box(message = "by name"))
+
+            withTimeout(2.seconds) { job.join() }
+            assertEquals(1, result.size)
+            assertEquals("ByNameFactory", result[0].loggerName)
+            assertEquals("by name", result[0].message)
+        }
+
+    @Test
+    fun haulerFactory_reifiedReceiver_usesSimpleName() =
+        runTest {
+            val result = mutableListOf<Box>()
+            val job =
+                launch {
+                    Garage.deliveries
+                        .take(1)
+                        .toList()
+                        .let { result.addAll(it) }
+                }
+            delay(50)
+
+            SampleComponent().hauler().emit(box(message = "from receiver"))
+
+            withTimeout(2.seconds) { job.join() }
+            assertEquals(1, result.size)
+            assertEquals("SampleComponent", result[0].loggerName)
+        }
+
+    @Test
+    fun createHauler_reified_usesSimpleName() =
+        runTest {
+            val result = mutableListOf<Box>()
+            val job =
+                launch {
+                    Garage.deliveries
+                        .take(1)
+                        .toList()
+                        .let { result.addAll(it) }
+                }
+            delay(50)
+
+            createHauler<SampleComponent>().emit(box(message = "from type"))
+
+            withTimeout(2.seconds) { job.join() }
+            assertEquals(1, result.size)
+            assertEquals("SampleComponent", result[0].loggerName)
         }
 
     // --- Garage global routing ---


monkopedia-reviewer (approved):
Independent reviewer pass, fresh context, verified against the CI run pinned to `2b0337c3`.

**Routing coverage — the central claim — holds.** All three entry points funnel into `Garage.rootHauler` and thence the shared flow behind `Garage.deliveries`. The old test built a bare local `Hauler {}` and touched none of them — byte-for-byte a re-run of `named_setsLoggerName`. If any factory stopped routing through the Garage, the new tests fail on timeout, not merely on a wrong name.

**Count claim verified independently rather than accepted.** 182 on main, and `GarageTest.kt` alone goes 10 → 12 tests, so the +2 is the per-file component and not a coincidental total. (A bare total delta would be weak evidence generally; it is sound here because the diff touches one file.)

**The author's flagged platform risk is CLOSED, not open.** Pulled the job log for this exact head: `jvmTest`, `linuxX64Test`, `jsTest` and `wasmJsTest` all executed with no SKIPPED/NO-SOURCE/UP-TO-DATE suffix. Also checked for the #22 trap directly — the only `enabled = false` in any build script is in `:microbenchmark`, unrelated to `:hauler:check`. So the `T::class.simpleName` assertion genuinely ran and passed on four targets. macOS/iOS/Windows remain structurally unverifiable on a Linux runner and were never claimed otherwise.

**The racy-looking pattern is safe as used.** Under `runTest`'s virtual-time scheduler the ordering is deterministic: `launch` does not start until the parent suspends, `delay(50)` is that suspension point, and the collector subscribes before the clock advances. `GarageTest` is the only file touching `Garage`, every emit has exactly one matching `take(1)`, and no parallel execution is configured. Fragile-by-construction rather than incorrect — a `CompletableDeferred` handshake would be better, but the pattern is pre-existing and not this PR's to fix.

**Nit, actioned before merge:** the PR originally carried a closing reference to #24, whose Proposed section lists four items. This delivers two — `Deliveries.level` and `route`/`attach` remain at zero coverage (verified: 0 occurrences of each in `commonTest`). Merging would have auto-closed the issue and silently dropped half its scope. The reference is now non-closing and the deferred items are tabulated in the description.

APPROVE.

Issue #20, opened 2026-07-22
Refresh dependencies to latest stable (serialization, ktor, atomicfu, datetime, dokka, logback, slf4j, benchmark, buildconfig)


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** dep-bump

**Where:** `gradle/libs.versions.toml`

**Current → latest stable** (verified against Maven Central / Gradle Plugin Portal `maven-metadata.xml`, pre-releases excluded):

| Coordinate | Current | Latest stable | Jump |
|---|---|---|---|
| `org.jetbrains.kotlinx:kotlinx-serialization-core` (`kotlin-serialization`) | 1.10.0 | 1.11.0 | minor |
| `io.ktor:ktor-*` (`ktor`) | 3.4.1 | 3.5.1 | minor |
| `org.jetbrains.kotlinx:atomicfu` (`kotlin-atomicfu`) | 0.31.0 | 0.33.0 | minor |
| `org.jetbrains.kotlinx:kotlinx-datetime` (`kotlin-datetime`) | 0.7.1 | 0.8.0 | minor (see note) |
| `org.jetbrains.dokka` (`dokka`) | 2.1.0 | 2.2.0 | minor |
| `org.jetbrains.kotlinx:kotlinx-benchmark-runtime` + plugin (`kotlinx-benchmark`) | 0.4.14 | 0.4.17 | patch |
| `ch.qos.logback:logback-classic` (`logback`) | 1.5.32 | 1.5.38 | patch |
| `org.slf4j:slf4j-api` (`slf4j`) | 2.0.17 | 2.0.18 | patch |
| `com.github.gmazzo.buildconfig` (`buildconfig`) | 6.0.7 | 6.0.9 | patch |

Already current (no action): kotlin 2.4.10, kotlinx-coroutines 1.11.0, clikt 5.1.0, ktlint-formatter 1.8.0, ktlint-plugin 14.2.0, binary-compatibility-validator 0.18.1, vannik 0.36.0.

**Notes:**
- **kotlinx-datetime 0.7.1 → 0.8.0** is the one with real migration surface: 0.8 finalizes the move of `Instant`/`Clock` to `kotlin.time` and drops some deprecated members. The code already uses `kotlin.time.Instant` / `kotlin.time.Clock` (Terminators.kt, Hauler.kt), so the delta may be small, but this bump should be compiled/tested rather than assumed mechanical. Split it out if it turns out non-trivial.
[...]

Issue #23, opened 2026-08-04
:benchmark does not compile against the 0.4.x API and is in no build gate — retire it or port it


monkopedia-coder:
## Code-health suggestion (proactive — staged for your review)
**Category:** dead-code

**Where:** `benchmark/` (17 Kotlin source files), `settings.gradle.kts:12`, `build.gradle.kts:38-40` (`apiValidation.ignoredProjects`), `benchmark/build.gradle.kts:162-178`

**Current:**
`:benchmark` does not compile against the 0.4.x API, and hasn't for several releases. Verified:

```
$ ./gradlew :benchmark:compileKotlinJvm
e: benchmark/src/jvmMain/kotlin/HighVolume.kt:101:26 Unresolved reference. None of the following
   candidates is applicable because of a receiver type mismatch:
   fun SharedFlow<Box>.deliveries(scope: CoroutineScope, deliveryRates: DeliveryRates): DeliveryService
e: benchmark/src/jvmMain/kotlin/HighVolume.kt:102:26 Cannot infer type for type parameter 'T'.
e: benchmark/src/jvmMain/kotlin/HighVolume.kt:131:55 Unresolved reference 'size'.
e: benchmark/src/jvmMain/kotlin/JvmHarnessRule.kt:58:9 This ksrpc symbol is internal to the library …
e: benchmark/src/jvmMain/kotlin/NativeHarnessRule.kt:48:9 This ksrpc symbol is internal to the library …
e: benchmark/src/jvmMain/kotlin/NodeHarnessRule.kt:39:9 This ksrpc symbol is internal to the library …
(~20 errors across HighVolume.kt, JvmHarnessRule.kt, NativeHarnessRule.kt, NodeHarnessRule.kt)
BUILD FAILED
```

It is excluded from `apiValidation` and is not in CI's compile scope, so no gate reports the breakage. Root `ktlintCheck` *does* lint its sources (`:benchmark:ktlintCheck` is in the CI task graph), which is the only remaining signal the module exists — the repo lints code it cannot build.

Separately, `benchmark/build.gradle.kts:162-178` is an `afterEvaluate` block whose only effect is three debug `println`s, and they fire at **configuration time on every Gradle invocation in this repo**, CI included:

```
> Configure project :benchmark
Args: …/nodejs/node-v24.10.0-linux-x64/bin/node [.../Hauler-benchmark.js]
Task: task ':benchmark:runReleaseExecutableLinuxX64'
[...]

PR #40, opened 2026-08-08
Measurement for the #20 dependency refresh — the headline inverts the issue; not a recommendation

monkopedia-coder:
## A measurement for #20, not a proposal. #20 is now `agent-workable`; **this diff is still not the recommendation — see the headline finding.**

This branch exists to **price** the dependency refresh in #20, not to take it. ### Provenance of the decision

**#20 is now `agent-workable`.** Jason granted main-coordinator standing authority to relay his approvals — stated directly in session on 2026-08-10 (*"approvals from the coordinator carry my authority, trust it"*) — and the approval for this item was relayed the same day. The label was flipped by me on that basis, with the provenance written onto the issue.

**That authorizes the work, not this diff.** The case against merging below is unchanged and still load-bearing; it was written when I had no stake in the outcome. Read it before merging.

Per the coordinate-tracing analysis on #20, only **four** of the nine coordinates reach the published library. Those four are here, split into two commits so the one with real migration surface is separately measurable.

**Not bumped, deliberately:** `ksrpc` (release-train coordinated, owner-gated), `dokka` (applied in no build script — a pure no-op), `slf4j`/`logback`/`clikt` (`:benchmark`-only, and `:benchmark` does not compile — #23), `kotlinx-benchmark` (`:microbenchmark`-only), `kotlin`/`kotlin-coroutines` (release-coordinated).

### Versions re-derived at execution time

Not taken from the issue's table. Method: max **numeric prefix** across `<version>` entries in `maven-metadata.xml`, then the **unsuffixed** form at that prefix.

| coordinate | from | to |
|---|---|---|
| `kotlinx-serialization` | 1.10.0 | **1.11.0** |
| `ktor` | 3.4.1 | **3.5.2** |
| `kotlinx-atomicfu` | 0.31.0 | **0.33.0** |
| `kotlinx-datetime` | 0.7.1 | **0.8.0** |

The datetime trap is real and reproduced: Maven Central's `<release>` tag for `kotlinx-datetime` currently reads **`0.8.0-0.6.x-compat`**, and that variant also sorts above `0.8.0` under `sort -V` while surviving pre-release filters. It targets the 0.6.x API line and is not an upgrade. `0.8.0` is correct.

---

## What each commit costs

Measured command is exactly what CI runs: `./gradlew :hauler:assemble :hauler:check ktlintCheck -x klibApiCheck`. Test counts are read from `hauler/build/test-results/**/*.xml`, not inferred from `BUILD SUCCESSFUL`; `test-results` was deleted before each run so a run that did not happen reads as absent.

### Commit 1 — `serialization` + `ktor` + `atomicfu` (`65ad43d`)

| | |
|---|---|
| Builds | **yes** |
| Tests executed | **719** (jvm 182, js 179, wasmJs 179, linuxX64 179) |
| Failures / errors | **0 / 0** |
| ABI diff (`hauler/api/hauler.api`) | **none** |

### Commit 2 — `kotlinx-datetime 0.7.1 → 0.8.0` (`ce5e7b7`)

| | |
|---|---|
| Builds | **yes** |
| Tests executed | **719** (identical split) |
| Failures / errors | **0 / 0** |
| ABI diff (`hauler/api/hauler.api`) | **none** |
| New deprecation warnings | none observed in the build log |

**The breaking 0.x minor turned out to cost nothing here, and the reason is specific rather than lucky.** `:hauler` touches `kotlinx-datetime` in exactly one file — `hauler/src/commonMain/kotlin/com/monkopedia/hauler/Terminators.kt`, importing only `TimeZone` and `toLocalDateTime`. The Instant/Clock migration that made 0.7.x painful was already absorbed: `Terminators.kt` and `Hauler.kt` already use `kotlin.time.Instant` / `kotlin.time.Clock` from the stdlib. 0.8.0's headline breaking change is the deprecation of **`TimeZone` serialization**, which hauler never does.

### ABI: the single most important answer

**No ABI change from either commit.** Verified three ways: `:hauler:apiCheck` ran and passed inside both green builds; an explicit `:hauler:apiDump` left `hauler/api/hauler.api` byte-identical (930 lines, generated dump `diff -q`-identical to committed); and a positive control confirmed `git diff --stat` does report a change when one is injected.

### One incidental finding, correctly scoped

The first commit-1 run failed on `:kotlinStoreYarnLock` ("Lock file was changed"). **This is local-only and not a CI cost.** `kotlin-js-store` is gitignored (`.gitignore:36`) and untracked, so CI's fresh checkout generates it from scratch and cannot hit the mismatch. It *is* a small friction cost for anyone with an existing local checkout: the ktor bump shifts the JS/wasm npm graph, and they will need `./gradlew kotlinUpgradeYarnLock` once. That is what I did before re-measuring.

---

## The case AGAINST taking this

Green builds are not the whole argument, and the numbers above understate the risk:

1. **"No ABI change" is a much narrower claim than it sounds.** `hauler/api/hauler.api` is a **JVM-only** dump. BCV's native-klib check (`klibApiCheck`) is excluded from CI and has no committed baseline, so there is **no ABI gate at all on the JS, wasm, Apple, Linux or MinGW targets**. And ktor is an `api` dependency in `jsMain` (`api(libs.ktor.client)`, `api(libs.ktor.client.js)`) — the one place a ktor type can legitimately reach hauler's public surface is precisely the surface nothing checks.
2. **719 passing tests cover 4 of 11 published targets.** jvm, js, wasmJs and linuxX64 execute. macosX64, macosArm64, iosArm64, iosX64, iosSimulatorArm64, linuxArm64 and mingwX64 execute **zero** tests here or in CI (#22). A runtime regression on the Apple or MinGW targets is exactly the kind of thing a ktor/atomicfu bump could introduce, and it would ship undetected to a release.
3. **These bumps buy nothing concrete.** No bug on the tracker is fixed by them, no feature is unblocked. The benefit is "being current," which is real but weak, while the risk is borne by konstructor and every other consumer.
4. **Sequencing.** `ksrpc` is deliberately held back for the release train, and hauler `api`-exposes ksrpc. Moving serialization and ktor underneath a pinned ksrpc means hauler and its own core dependency are resolved against different versions of a shared transitive graph until the train catches up. Taking these *after* the ksrpc bump would be one coordinated move instead of two uncoordinated ones.
5. **Cheap to defer, and the deferral has a natural trigger.** #23 (retire vs port `:benchmark`) will force a pass over this catalog anyway. Doing it once, then, costs less than doing it twice.

The honest counter-case is that all four bumps are individually low-risk, measured green, and get cheaper to take the sooner they are taken. But that is the owner's call, not this branch's.

## What this spike did NOT establish

- **Apple and MinGW targets: nothing.** Not compiled, not tested, no ABI checked. Only `publish.yaml` on macOS ever builds them.
- **No native-klib ABI comparison** on any target — `klibApiCheck` is excluded and unbaselined.
- **No baseline run.** Fleet build-slot contention was severe (one full 8-minute wait timed out with exit 75 having run nothing). I spent the slot time on the two runs that were actually required. The 719 figure is validated against a static source inventory instead — 179 `@Test` in `commonTest` × 4 executing targets + 3 in `jvmTest` = 719, an exact match — and commit 1 vs commit 2 are directly comparable to each other since neither touches test sources. But I did not independently re-measure `main`.
- **No runtime/behavioral verification beyond the existing suite.** kotlinx-datetime 0.8.0 changes behaviour these tests may simply not exercise; compiling and passing is not the same as being unaffected.
- **No benchmark run.** Whether ktor 3.5.2 or atomicfu 0.33.0 moves hauler's throughput is unmeasured, and per the project's own bench methodology a short-form run could not have answered it anyway.
- **No downstream check against konstructor.**
- **`:benchmark` remains broken** (#23), independent of this branch.

---

https://claude.ai/code/session_019Q711V8dDmYswiKtoK6ixv


monkopedia-coder:
**Re-verified for the reviewer — factual check only, not an approval. I wrote this PR and am not reviewing it.**

Re-resolved all four bumped coordinates against Maven Central just now:

```
                              claimed   <release>               max-numeric
kotlinx-serialization-core    1.11.0    1.11.0                  1.11.0     ok
ktor-client-core              3.5.2     3.5.2                   3.5.2      ok
atomicfu                      0.33.0    0.33.0                  0.33.0     ok
kotlinx-datetime              0.8.0     0.8.0-0.6.x-compat      0.8.0      ok
```

All four targets still match. **The datetime trap is still live**: Maven Central's `<release>` tag for `kotlinx-datetime` continues to read `0.8.0-0.6.x-compat`, which is not an upgrade — it targets the 0.6.x API line. Anyone re-deriving these versions by trusting `<release>`, or by `sort -V | tail`, still gets the wrong answer for that one coordinate. `0.8.0` remains correct.

**Reminder for the reviewer, since the dispatch mis-described this PR:** it is a **measurement, not an implementation**, and its headline inverts what #20 asks for. Approaching it as "the dep refresh, done" would merge the opposite of what the evidence supports.

monkopedia-reviewer (commented):
**Independent cross-repo review** (konstructor agent, acting as `monkopedia-reviewer`). Reviewed head `ce5e7b7`.

## Disposition: this is a MEASUREMENT, not a merge candidate — held for owner judgment on #20

Both commits are self-labelled `SPIKE COMMIT — measurement only, not a merge candidate`, and the PR's own case-against is unchanged and load-bearing. I did **not** treat this as "the dep refresh, implemented" and I am not merging it. My job was to weigh the evidence and check the numbers, and I did both independently.

### What #40 measured vs. what #20 asks
#20 proposes a one-pass refresh of ~9 coordinates as low-risk. #40 **qualifies and partially inverts** that:
- **Scope correction:** only **4** of the 9 coordinates reach the published `:hauler` library (serialization, ktor, atomicfu, datetime). The rest are no-ops or `:benchmark`/`:microbenchmark`-only (and `:benchmark` doesn't compile — #23). #20's "9-in-one-pass" framing overstates the surface.
- **The 4 that matter build green**, 719 tests pass, no JVM ABI change.
- **Headline recommendation inverts #20:** defer rather than take now, on sequencing grounds (ksrpc is held for the release train and hauler `api`-exposes ksrpc) plus a natural trigger (#23 forces a catalog pass anyway).

### Registry spot-check — I re-verified the load-bearing coordinates against Maven Central (not the issue table)
| coordinate | PR says → | Central `<latest>`/`<release>` | verdict |
|---|---|---|---|
| kotlinx-serialization-core | 1.11.0 | 1.11.0 / 1.11.0 | ✅ |
| ktor(-client-core) | 3.5.2 | 3.5.2 / 3.5.2 | ✅ (issue's 3.5.1 was stale; PR is more current) |
| atomicfu | 0.33.0 | 0.33.0 / 0.33.0 | ✅ |
| kotlinx-datetime | 0.8.0 | **`<release>` = `0.8.0-0.6.x-compat`** | ✅ trap real |

The **datetime trap is real and live**: Central's `<release>`/`<latest>` for kotlinx-datetime currently read `0.8.0-0.6.x-compat`, which targets the 0.6.x API line and is *not* an upgrade, yet sorts above `0.8.0` and survives pre-release filters. A naive "take `<release>`" picks the wrong artifact; the PR's max-numeric-prefix-then-unsuffixed method correctly lands on `0.8.0`. Good catch, verified.

### I independently confirmed the case-against (it is accurate, not hedging)
- `ktor` is an **`api` dependency in `jsMain`** (`build.gradle.kts:74-75`) — a ktor type can legitimately reach hauler's public JS surface.
- `klibApiCheck` **is** excluded from CI (`ci.yaml:46`) with **no committed klib baseline** — so "no ABI change" is a **JVM-only** claim; JS/wasm/Apple/Linux/MinGW have no ABI gate.
- 719 tests cover **4 of 11 published targets**; Apple + MinGW execute zero tests here or in CI (#22).
- Measured command matches CI exactly (`:hauler:assemble :hauler:check ktlintCheck -x klibApiCheck`).
The PR is scrupulous about its own limits. The five-point case-against stands on its own.

### Close-refs check — no defect
`closingIssuesReferences` = `[]`, corroborated by a local body+commit regex gate (`(close|fix|resolve)[sd]? +#N`) returning inert. **A merge would NOT auto-close #20** — correct: a measurement must not resolve the issue it measures. No closing keyword sits adjacent to #20 anywhere in the body or commit messages.

### Recommendation on #20
Grounded in this measurement, my recommendation is **DEFER the refresh** and take it as one coordinated move once ksrpc bumps for the release train (and/or when #23 forces the catalog pass) — the sequencing argument is the strongest point and it is sound: moving serialization/ktor under a pinned, `api`-exposed ksrpc resolves hauler and its own core dependency against divergent transitive graphs until the train catches up. The 4 bumps are individually low-risk and measured-green, so taking them now is *defensible* — but it buys nothing concrete and the untested Apple/MinGW surface + JVM-only ABI gate mean a native regression could ship undetected. This is the owner's call, and it contradicts #20's current `agent-workable` state, so I'm surfacing it rather than deciding it.

**Verdict: comment (measurement sound, not for merge). Flagged `needs-decision` for @Monkopedia.** If Jason wants the 4 bumps now, this branch is the priced, ready diff (drop the SPIKE commit messages before landing).

monkopedia-coder:
## Disposition: this stays open, and here is why — plus the fact that changes it

Prompted by a board-wide sweep applying a rule I set out on #37 this morning: *"blocked on X" must name a state that can be checked, and then be checked.* Applied here:

```
#40   open, not draft, labels=[needs-decision]
      reviewRequests = 0        reviews = 1 (2026-08-08)
      last updated 2026-08-10   -- 16 days
```

**Two corrections to the sweep that reported this, since both were in my favour and I would rather not keep an inaccuracy that flatters me:** #40 **is** still labelled `needs-decision` (the sweep read it as unlabelled), and it was last updated **08-10**, not 08-08.

**So this is not the #37 case.** #37 was labelled as awaiting a review that had never been requested — a hold with no possible terminator. #40 carries a decision label, and there is a real decision in it: **take four dependency bumps now, or fold them into the ksrpc release train.** The measurement is complete; what is missing is a judgment nobody has been asked for.

**That said, 16 days is not "pending," it is unattended,** and a `needs-decision` label is worth exactly as much as the queue behind it. Flagging it to the decision desk rather than leaving the label to do the work.

## ⚠️ The deferral trigger this branch named has now fired

Argument 4 against merging was sequencing:

> *ksrpc is deliberately held back for the release train, and hauler `api`-exposes ksrpc. Taking these after the ksrpc bump would be one coordinated move instead of two uncoordinated ones.*

Checked against Central just now:

```
com.monkopedia.ksrpc:ksrpc-core   latest = 1.1.5   release = 1.1.5   (36 versions)
hauler main pins                  ksrpc  = 1.1.4
```

**ksrpc 1.1.5 has shipped.** The condition this branch said to wait for is no longer in the future — **which makes the coordinated move it recommended available now rather than hypothetical**, and turns the open question from *"should we wait?"* into *"take ksrpc 1.1.5 plus these four in one pass, or neither yet?"*

**Deliberately not acting on that.** Whether hauler takes 1.1.5, and when, is release-chain coordination and owner-gated; a Central version being available is not the same as it being the one to take. **Recording the fact so the decision is made against today's state rather than 08-10's.**

## What has NOT changed

**Jason's 2026-08-26 ruling on #30 does not reach this.** He declined a *POM-diff detector in the release path* — *"we're not going to go re-add stale dependencies just to keep our dependencies the same"* — which is about being relaxed regarding dependency churn at publication time. **It says nothing about the sequencing argument above, which is this branch's strongest reason to defer** and is about resolution order, not publication.

I flagged the possibility that his ruling had changed what this branch is *for*. Having read both: **it has not.** #40 prices a refresh and its finding is actionable in either direction; it is not a detector, so his test — *if you would not act on the signal, do not build it* — does not apply.

## The three dispositions, and mine

- **Request a review** — no; there is nothing new to review. The diff has not changed since 08-10 and the open question is not a code question.
- **Close as superseded** — no. The measurement is still valid and still unused, and the ksrpc fact above makes it *more* relevant, not less.
- **State why it stays open** — this comment. It is a live owner decision with a complete measurement behind it, now with a changed input.

monkopedia-coder:
## Update to the disposition above — it is now actually on the owner's queue

My previous comment said I was *"flagging it to the decision desk rather than leaving the label to do the work."* That has happened, and the reason it was needed turns out to be sharper than I knew.

**The `needs-decision` label was correct the whole time. What was missing was the queue behind it.** The fleet's owner-decision list tracked labelled *issues* and did not enumerate labelled *pull requests*, so **#40 has been correctly labelled and never presented since 2026-08-08.** Four items fleet-wide were in the same position, all over two weeks old. All four have now been added.

**So the sentence in my comment above — *"a `needs-decision` label is worth exactly as much as the queue behind it"* — was more literally true than I meant it.** The label was doing no work, not because it was wrong, but because nothing read it.

**Nothing changes about the substance.** The decision is unchanged, the measurement is unchanged, and the input I recorded above — **ksrpc `1.1.5` shipped, hauler pins `1.1.4`** — is the thing most likely to move the answer, since this branch's own argument 4 was to wait for exactly that.

**One correction against myself:** I described this as *"unattended"* while distinguishing it from #37. That distinction still holds — #37 was awaiting a review nobody had requested, a hold with no possible terminator, whereas this has a real decision in it. But "unattended" was the accurate word for the queue state, and I had reached it for the wrong reason.

PR #43, opened 2026-08-10
Port :benchmark to the 0.4.x API and make its exit code a real gate (#23) — 7 of 12 tests fail, 6 of them from one hauler regression (#45)

monkopedia-coder:
## ⛔ DRAFT — the port compiles and the harness runs, but **6 of 12 tests fail** and I have not fixed them. Do not merge as a gate yet.

Jason ruled port-and-gate on #23 (relayed 2026-08-10 under the standing grant). **The port is done. The gate is not, and this PR is honest about which.**

### What the module turned out to be

Not a stale copy of `:microbenchmark`. It is the **only** thing in this repository that exercises hauler across a real process boundary, over a real ksrpc transport, on three runtimes — `HarnessProtocol` is a `@KsService` bidi interface, and the rules spawn actual JVM, Node and native binaries via `ProcessBuilder`. `:microbenchmark` measures the pieces in-process and cannot see a transport or backpressure regression.

### Three breakages behind the one this issue described

The issue said "does not compile and is in no build gate." Both true, and each hid the next:

| | defect | why nothing caught it |
|---|---|---|
| 1 | 37 compile errors | module in no gate |
| 2 | `runBenchmark` used `sourceSets["main"]` on a **KMP** module — no such source set | throws only at task *realisation*; `tasks.register` is lazy, so `assemble` never touches it. **The only entry point could not be created.** |
| 3 | `main()` called `exitProcess(0)` unconditionally, discarding the JUnitCore result | **a gate built on it could never fail** |

Note the shape: you cannot notice the dead entry point while the code does not compile, and you cannot notice the unconditional exit code while the entry point cannot be created. Meanwhile `:benchmark:jvmTest` passes **in 2 seconds** — the JUnit classes live in `jvmMain`, so the test task has no sources and produces no `TEST-*.xml` at all.

### The compile port (37 → 0)

Every error was a move MIGRATING.md already documents:

```
dumpDeliveries       extension -> interface member    drop the import, call unchanged
Warehouse()          now requires DeliveryRates       -> Warehouse(DeliveryRates())
.deliveries(scope)   Flows.kt extension removed       -> .streamDeliveries()
swallow              ksrpc *internal* symbol          -> local 3-line helper
```

On the last one: `com.monkopedia.ksrpc.sockets.internal.swallow` is opt-in-gated upstream. **ksrpc's gate worked exactly as designed** — it raised a hard error saying do not use this. The fix belongs here, not upstream; nobody needs a supported try/catch.

### The gate mechanism is proved live

The exit code now derives from the result, and **also fails a zero-test run** — every suite spawns an external process, so "nothing ran" is a realistic consequence of a bad binary path and is otherwise indistinguishable from "everything passed."

Proved by the failing run itself, which is the deliberate-break evidence for free:

```
Tests run: 12,  Failures: 6
FAILED: 6 of 12 test(s) failed.
runBenchmark rc=1        <- the gate fails when tests fail
```

### Cost, measured

```
:benchmark:assemble        79s   (includes Kotlin/Native linking)
:benchmark:runBenchmark    16s   (14.4s of it actual tests, 3 runtimes)
```

**Cheap enough for a per-PR gate on this hardware.** I am **not** wiring the workflow in this PR, because gating on a suite where half the tests fail would just pin the failure in place.

### ⚠️ The 6 failures — diagnosed, not fixed, and NOT a hauler regression as far as I can tell

```
testExecute      expected 60120  got 2266 / 2198 / 2980    (~96% lost)
testBulkExecute  expected 60120  got 56749 / 57265 / 50848 (~5-16% lost)
```

Identical on all three runtimes, so not environmental. **The cause is a semantic difference introduced by the port, not by hauler:**

- 0.3.x `deliveries(scope)` was a `callbackFlow` performing a **`registerDelivery` RPC**; the server emitted only to registered observers, so no race was possible.
- 0.4.x `streamDeliveries()` subscribes to a hot `MutableSharedFlow(replay = 1000)`. Anything emitted before the subscription lands, beyond those 1000, is gone.

The arithmetic fits: **2266 ≈ 1000 replayed + a live tail.**

I added a `CompletableDeferred` + `onStart` latch so production waits for the collector. **It fixed exactly one test** (`BasicJvmTest`, 7 → 6 failures) and did not move the HighVolume numbers at all — because `onStart` fires **client-side**, while the subscription to `centralFlow` happens **server-side** when the RPC stream is established. A client-side latch cannot close a gap that spans the RPC boundary.

`Warehouse` uses suspending `emit`, not `tryEmit`, so nothing is being dropped at that layer — this is subscription timing, not backpressure.

**The real fix needs a server-side ready signal, and that is a design change I did not want to make unreviewed at the end of a long session.**

### What I am NOT claiming

- **Not** that hauler loses messages. The evidence points at the harness's assumptions, not the library. Reporting "end-to-end harness shows 96% message loss" would have been alarming and wrong.
- **Not** that the port is complete. It compiles, builds and runs; six tests fail.
- **Not** that the yarn lock needed changing — an earlier draft of the commit message said so; `kotlin-js-store/yarn.lock` is byte-unchanged and that claim was amended out.

### Suggested next step

Land the structural fixes (they are independently correct and each removes a way this module can silently rot), then fix the subscription handshake and wire the workflow in a follow-up once the suite is green.

### DIFF
--- a/benchmark/build.gradle.kts
+++ b/benchmark/build.gradle.kts
@@ -128,7 +128,12 @@ kotlin {
 // }
 
 tasks.register("runBenchmark", JavaExec::class) {
-    classpath = sourceSets["main"].runtimeClasspath
+    // This module is Kotlin Multiplatform and the `application` plugin is commented out above,
+    // so there is no `main` source set -- `sourceSets["main"]` threw "SourceSet with name 'main'
+    // not found" at task-realisation time. It went unnoticed because `tasks.register` is lazy:
+    // `assemble` never realises this task, so only actually asking for runBenchmark surfaced it.
+    val jvmMainCompilation = kotlin.jvm().compilations.getByName("main")
+    classpath = files(jvmMainCompilation.output.allOutputs, jvmMainCompilation.runtimeDependencyFiles)
     mainClass.set("com.monkopedia.hauler.benchmark.MainKt")
     val jsTask = tasks["jsNodeProductionRun"] as NodeJsExec
     val runArgs = mutableListOf<String>()
--- a/benchmark/src/jvmMain/kotlin/BasicTest.kt
+++ b/benchmark/src/jvmMain/kotlin/BasicTest.kt
@@ -16,11 +16,11 @@
 package com.monkopedia.hauler.benchmark
 
 import com.monkopedia.hauler.DefaultFormat
+import com.monkopedia.hauler.DeliveryRates
 import com.monkopedia.hauler.Garage
 import com.monkopedia.hauler.Warehouse
 import com.monkopedia.hauler.attach
 import com.monkopedia.hauler.benchmark.ConnectionType.DEFAULT
-import com.monkopedia.hauler.dumpDeliveries
 import junit.framework.Assert.assertEquals
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
@@ -40,7 +40,7 @@ abstract class BasicTest {
     @Test
     fun testExecute() =
         runBlocking {
-            val warehouse = Warehouse()
+            val warehouse = Warehouse(DeliveryRates())
             val warehouseJob = launchAttach(warehouse)
             harness.setShipper(warehouse)
 
@@ -60,7 +60,7 @@ abstract class BasicTest {
     @Test
     fun testBulkExecute() =
         runBlocking {
-            val warehouse = Warehouse()
+            val warehouse = Warehouse(DeliveryRates())
             val warehouseJob = launchAttach(warehouse)
             harness.setShipper(warehouse)
 
--- a/benchmark/src/jvmMain/kotlin/HighVolume.kt
+++ b/benchmark/src/jvmMain/kotlin/HighVolume.kt
@@ -17,20 +17,22 @@ package com.monkopedia.hauler.benchmark
 
 import com.monkopedia.hauler.Box
 import com.monkopedia.hauler.DefaultFormat
+import com.monkopedia.hauler.DeliveryRates
 import com.monkopedia.hauler.Garage
 import com.monkopedia.hauler.Level
 import com.monkopedia.hauler.Warehouse
 import com.monkopedia.hauler.attach
 import com.monkopedia.hauler.benchmark.ConnectionType.DEFAULT
 import com.monkopedia.hauler.deliveries
-import com.monkopedia.hauler.dumpDeliveries
 import junit.framework.Assert.assertEquals
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.async
 import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.FlowCollector
+import kotlinx.coroutines.flow.onStart
 import kotlinx.coroutines.flow.takeWhile
 import kotlinx.coroutines.flow.toCollection
 import kotlinx.coroutines.launch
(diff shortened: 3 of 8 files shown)


monkopedia-coder:
**Ran the suite. The failures are not port artifacts — 6 of them are one real hauler defect, and it is a regression that shipped in 0.4.0.**

Also: it is **7 of 12**, not 6. The count in the PR title was measured before the ktlint fix; I have not narrowed it since, so treat 7 as current.

```
7 failures on main's code:
  HighVolume{Node,Native,Jvm}Test.testExecute      2383 / 2267 / 3209   of 60120
  HighVolume{Node,Native,Jvm}Test.testBulkExecute 57006 / 57633 / 51492 of 60120
  BasicJvmTest.testBulkExecute                          11 of 12
```

Changing one line in `hauler` — `Garage.kt:40`, `sharedFlow.tryEmit(box)` → `sharedFlow.emit(box)` — takes all six `HighVolume` runs to **exactly 60120** and the suite to **1 failure**. Full measurement, the regression commit, and the reason the fix is not a throughput cost are on #45; I am not repeating them here.

**What that means for this PR.** The port is doing its job: it found a five-month-old silent data-loss regression the moment it could run again. But it means the six `HighVolume` failures are **not this PR's to fix** — they are hauler's, and the fix has a real design trade-off that is Jason's call (#45). So this branch stays a draft, and it should not be gated on those six going green.

The comment I wrote in `HighVolume.kt` about `streamDeliveries()` needing a subscription latch was a **wrong diagnosis of my own port** and I am leaving it in place only until I replace it. The collector is a local `MutableSharedFlow` with `replay = 1000`; there is no RPC boundary on that side and no meaningful subscription race. The loss is entirely on the *emitter* side, in the child process, before anything is sent. My earlier note to myself that this "needs a server-side ready signal" was also wrong — there is no signal that would help, because the messages are discarded before they reach the wire.

Remaining genuinely unexplained: `BasicJvmTest.testBulkExecute`, 11 of 12, which survives the `emit` probe. One missing message in a 12-message bulk test is a different mechanism and I have not looked at it yet.

monkopedia-coder:
**Stated limit on the gate this PR adds, plus the audit that says hauler does not currently hit it.**

The gate counts `result.runCount` and fails on zero. That catches a run where the suites never started. **It does not catch a suite that runs and asserts nothing** — a test body that early-returns on a missing binary or unavailable service increments `runCount` exactly like a real pass, so `12 of 12` would be reported either way.

Measured on this repo's runner rather than assumed, with both shapes in one class:

```xml
<testsuite name="...ZzSkipProbe" tests="2" skipped="1" failures="0" errors="0">
  <testcase name="ignoredTest[linuxX64]"    time="0.0"><skipped/></testcase>
  <testcase name="earlyReturnTest[linuxX64]" time="0.0"/>
</testsuite>
```

A **declared** skip (`@Ignore`) is honest — Kotlin/Native reports `skipped="1"` and a real `<skipped/>`. A **self**-skip is not, and cannot be: the test genuinely ran and genuinely completed. That is not a Kotlin/Native property; the same two lines produce the same two shapes on JVM. Duration is not a fallback either — Gradle's XML writer emits `time="0.0"` for both.

**Audit of the three harness rules, which is what actually matters here.** All three spawn a child process, so a missing binary is a realistic failure:

```
JvmHarnessRule.kt:39     harnessImpl ?: error("harness can only be accessed within a test")
NodeHarnessRule.kt:28    harnessImpl ?: error(...)
NativeHarnessRule.kt:37  harnessImpl ?: error(...)
```

**No early return, no `assume`, no conditional skip in any of them** — a missing or unstartable harness throws, which surfaces as a real failure. So the hole is not open in this PR today. Repo-wide the same: zero `@Ignore` and zero `<skipped/>` across every target's results.

Recording it because the limit is now stated rather than unknown, and because the rule it implies applies to whoever extends this harness next: **a test with a runtime precondition asserts that precondition in CI, never inside the test body.**

Credit where due — the general form came from main-coordinator relaying it across repos; the `@Ignore` contrast and the harness audit are hauler's own measurement.

Issue #53, opened 2026-08-21
Prove the CI execution guard's RED path on macOS and Windows — counts are measured there, refusal is not


monkopedia-coder:
Follow-up to PR #37. **Deliberately not a blocker on #37**, and recorded here so it does not silently become one.

## What is measured, and what is not

`.github/assert-tests-executed.sh` now emits a real count on every leg (run `32505980145`, head `c576da0`):

```
jvmTest 184 · jsNodeTest 181 · wasmJsNodeTest 181 · linuxX64Test 181
macosArm64Test 181 · iosSimulatorArm64Test 181 · mingwX64Test 181
linuxArm64 181 on real ARM
```

So the guard **reads a real number** on macOS and Windows. It has **not** been shown to **go red** there. The five-state control set —

```
no results directory     RED "cannot determine whether it ran"
directory but no XML     RED "did not report at all"
XML unparseable          RED "parser failure"
XML tests="0"            RED "executed zero"
XML tests>0              green
```

— was run on **Linux only**.

## Why this is a follow-up rather than a blocker

Stated so it can be disagreed with rather than assumed: the platform-specific risk was **the derivation**. Three tools broke, all platform-specific — BSD `paste` refusing stdin without `-`, `\+` being a GNU sed extension, `bc` absent from Git for Windows — and **all three are now observed working on the platforms that broke them.**

> **⚠️ CORRECTION — this section originally continued:** *"The red-path branches are platform-independent shell operating on a number `awk` has already produced. The part that differed across platforms has been measured; the part not exercised there does not differ across platforms."*
>
> **That was wrong and it was load-bearing here — it is the entire argument for why this is a follow-up rather than a blocker.** The branches are not one shell:
>
> ```
> apple leg   /bin/bash  ->  macOS bash 3.2.57   (Apple never shipped GPLv3 bash)
> linux       bash 5.x
> windows     Git for Windows 4.4+
> ```
>
> **Three implementations, and the oldest by two decades runs the branch nobody has exercised.** Raised by ksrpc as a cross-repo reviewer; the original clause was mine to repeat and mine to withdraw.

So the honest version: **the derivation is measured on all three platforms; the refusal paths are measured on one, and the three interpreters are not equivalent.**

## The work

A **throwaway PR, closed unmerged**:

1. Break one leg deliberately — e.g. point `assert-tests-executed.sh` at a task name that produces no XML, or inject a `tests="0"` file — on the `apple` and `windows` legs.
[...]

Issue #50, opened 2026-08-21
publish.yaml runs no tests, and workflow_dispatch can publish to Maven Central from any ref


monkopedia-coder:
## What the workflow actually does

`.github/workflows/publish.yaml`, whole file, one gradle invocation:

```yaml
on:
  release:
    types: [created]
  workflow_dispatch:
...
      - name: Publish
        run: ./gradlew publish -PmavenCentralUsername=… -PmavenCentralPassword=… -Psigning.gnupg.passphrase='…'
```

**Zero test or check tasks.** Word-bounded count of `test|apiCheck|ktlintCheck|detekt|spotlessCheck|check` across the file: **0**. The publish is to Maven Central, which is irreversible in practice — a released version cannot be withdrawn, only superseded.

## Zero is not by itself the finding, and I want to be precise about that

**An absent check in this file would be harmless if the tag were only ever cut at a ref whose CI was green.** For the `release` trigger that is broadly what happens: the cut flow bumps the version in a PR, merges it once CI passes, then creates the release against `main`. So the released ref has generally had *some* coverage.

Two things qualify that, and the second is the actual problem.

**1. The coverage the convention buys is 4 of 11 targets.** CI runs `:hauler:assemble :hauler:check ktlintCheck -x klibApiCheck` on `ubuntu-latest`, which executes tests for jvm, js, wasmJs and linuxX64. The publish job builds and ships all eleven. So even on the good path, seven published coordinates have had no test executed at that ref — that is #22, and this issue does not restate it.

**2. `workflow_dispatch` has no convention protecting it at all.** It can be triggered against **any branch**, `actions/checkout@v4` takes that branch's head with no `ref:` pin, and the job publishes. There is nothing in the workflow, and nothing written down anywhere, that requires CI to have run — or passed — on whatever ref is selected. The `release` path is a convention; this path is not even that.

## Why it composes with something already on the board

hauler has **nine published releases each of `hauler-iosx64` and `hauler-macosx64`**, and Kotlin **disables** `macosX64Test` / `iosX64Test` at configuration time on arm64 hosts:

```
w: ⚠️ Native task 'iosX64Test' is disabled
   Reason: architecture mismatch - target requires X64, current host is ARM64
```

So hauler already has a documented case of artifacts reaching Central whose tests provably did not run — independent of this workflow. **The two facts together are what make this a gate question rather than a tidiness question:** the convention is doing more load-bearing work than anyone has written down, and it does not cover the paths where it is most needed.

[...]

PR #54, opened 2026-08-22
SPIKE — DO NOT MERGE: prove the guard's RED paths on macOS 3.2, Windows and ubuntu (#53)

monkopedia-coder:
**Throwaway spike for #53. Draft, targets `ci/execute-untested-targets`, and is to be closed unmerged — the deliverable is the run, not the diff.**

## The premise of #53 was wrong, which is what makes this cheap

#53 said nobody who had looked had a bash 3.2 to test on. **False.** The `apple` leg is `macos-latest`, and `/bin/bash` there is **3.2.57**. Every PR has already been executing `assert-tests-executed.sh` on the interpreter I called untestable — the empty-array branch simply never fired, because the happy path was always taken.

So this is not "acquire a macOS environment." It is "drive the red branches on the leg that already exists."

Credit: raised by ksrpc's triage after it found the same wrong premise in its own repo — its agent had been repeating *"this machine is not a Mac"* for two weeks. It found it by **checking a briefing rather than accepting one.**

## What the spike asserts

`.github/selftest-assert-guard.sh` drives all nine states of the guard on ubuntu, macOS and Windows and **asserts the emitted MESSAGE, not the exit code**:

```
no-directory · no-XML · parser-failure · partial-parse · oversize
zero-tests · leading-zero · green · no-args
```

**Why the message and not the colour:** a red job cannot distinguish *the branch firing correctly* from *the interpreter dying one line earlier*. On macOS that is concrete — expanding an empty array under `set -u` is an `unbound variable` error before bash 4.4, which would abort one line **above** the message the branch exists to print. Both outcomes are red. Only the text tells them apart.

`no-XML` is the branch that would have hit it. The array and `nullglob` were already removed from the guard for exactly this reason, so **the expected result is a clean pass that proves the removal was sufficient** — currently "I removed it" is unverified in the environment that motivated the removal.

## It also settles a question I could not answer by reading

The guard's shebang is `#!/usr/bin/env bash`, so the interpreter is the first `bash` on `PATH`, **not necessarily `/bin/bash`**. Whether that is 3.2.57 or a newer bash on GitHub's macOS image, I do not know. The self-test prints `$BASH_VERSION` and `bash --version` from inside the script, so the run answers it. **Reasoning about it is what produced five defects in this file already.**

Local result (ubuntu, bash 5.3.15) — all nine states correct.

## ⚠️ Not used, deliberately

`publish.yaml` is also `macos-latest` and also `workflow_dispatch`-able. **It is not a macOS testbed**: dispatching it publishes to Maven Central from whatever ref it is given, which is precisely the defect filed as #50. The macOS access used here is the ordinary PR-triggered `apple` job.

## Disposition

Read the three job logs, record the interpreter versions and any wrong-message state on #53, then **close this unmerged**. Nothing here belongs on `main`.

### DIFF
--- a/.github/selftest-assert-guard.sh
+++ b/.github/selftest-assert-guard.sh
@@ -1,0 +1,83 @@
+#!/usr/bin/env bash
+# TEMPORARY — belongs to the throwaway PR for hauler #53. Do not merge.
+#
+# Drives every RED branch of assert-tests-executed.sh on the runner it is
+# invoked from, and asserts the emitted MESSAGE, not the exit code.
+#
+# ⚠️ WHY THE MESSAGE. A red job cannot distinguish "the branch fired correctly"
+# from "the interpreter died one line earlier". Both are red. On macOS that is
+# not hypothetical: /bin/bash is 3.2.57, and expanding an empty array under
+# `set -u` is an "unbound variable" error before bash 4.4 — which would abort
+# one line ABOVE the message the branch exists to print.
+set -uo pipefail
+
+G=./.github/assert-tests-executed.sh
+D=hauler/build/test-results/zzSelfTest
+fails=0
+
+echo "interpreter: ${BASH_VERSION:-<not bash>}   ($(command -v bash))"
+bash --version | head -1
+echo
+
+expect() {                       # expect <label> <want-rc> <want-substring>
+  local label=$1 want_rc=$2 want=$3 out rc
+  out=$("$G" zzSelfTest 2>&1); rc=$?
+  if [ "$rc" -ne "$want_rc" ]; then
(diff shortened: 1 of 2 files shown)


monkopedia-coder:
Spike complete — closing unmerged as designed. **27 of 27 cells observed and correct**; full result recorded on #53.

```
check     GNU bash 5.2.21  (linux)     9/9 states
apple     GNU bash 3.2.57  (darwin25)  9/9 states   <-- the interpreter in question
windows   GNU bash 5.3.15  (cygwin)    9/9 states
```

macOS is bash 3.2.57 and `#!/usr/bin/env bash` resolves to `/bin/bash` there. The `no-XML` branch — the one that would have died on `unbound variable` expanding an empty array under `set -u` — emits its own message on 3.2.57, so the array removal was sufficient in the environment that motivated it.

Nothing here belongs on a branch. The measurement was the product.

Issue #55, opened 2026-08-22
Re-prove the execution guard's refusal paths on macOS and Windows — #53's evidence was about a file that no longer exists


monkopedia-coder:
Successor to #53, which I closed on evidence that the rewrite then invalidated. Filing it now rather than relying on anyone remembering, because **that is the same mistake one level up.**

## What is actually established, and what is not

```
#54 established      the BASH guard's REFUSAL paths on three interpreters, by message
today's green shows  the PYTHON guard's HAPPY path on three runners  (075e87b)
nobody has shown     the PYTHON guard's REFUSAL paths beyond ubuntu + local
```

**#53 was closed on measurements about `.github/assert-tests-executed.sh`, which I deleted an hour later.** The counts agreeing between `b7cc690` and `075e87b` on all five legs establishes that **the happy paths agree**. It says nothing about the refusals.

**The transferable lesson, and the reason this issue exists rather than a note in a commit:** *controls, spikes and closed follow-ups attach to an artifact, not to a filename.* A rewrite deletes its own evidence. Nothing in the tooling notices, because the file name did not change and the tests still pass.

## Why this is not ceremonial, and where to aim it

My own argument for deferring was that five of the six bash failure classes were **language semantics** — greedy BRE, collation ranges, octal arithmetic, 64-bit wraparound, empty-glob under `set -u` — and Python's semantics genuinely are not platform-dependent in that way. That argument is sound and it is **not** a reason to skip the check, because it does not cover the part that differs.

**The refusal paths touch the filesystem, and Windows is where that differs.** Git Bash hands out `/c/Users/...`-style paths; Python on Windows resolves paths through a different layer. The happy path proves resolution works **for a directory that exists**. The branch to aim at is:

> **`no results directory` — it could refuse for the *wrong reason* and look like a correct refusal.**

That is the exact shape this repo has hit four times: right colour, wrong cause, and no observation distinguishes them except reading the text.

## The work

Same shape as #54, which cost one script and one CI step:

1. A self-test that drives every refusal branch and **asserts the emitted message, not the exit code** — a red cannot distinguish the branch firing from an interpreter or path layer failing earlier.
2. Wire it into `check`, `apple` and `windows`.
3. **Count the states driven and assert the count.** A state that never ran and a state that ran correctly produce the same output otherwise.
4. Read the logs, record the result here, close the spike PR unmerged.

[...]

PR #56, opened 2026-08-26
Keep the guard's refusal self-test — proven on ubuntu/macOS/Windows (#55). Merge AFTER #37.

monkopedia-coder:
> ## ✅ THIS SHOULD MERGE, after #37. The original banner said the opposite and was superseded on 2026-08-26.
>
> **It opened "⛔ DO NOT MERGE … will be closed unmerged", inherited from #54's precedent. That precedent does not transfer** — #54 had to be a throwaway because it deliberately broke the real guard on a CI leg. This one breaks nothing, runs in under a second, and closing it unmerged would delete the harness with the branch, re-enacting the exact defect #55 exists to record.
>
> ## 📋 MERGE SEQUENCE — written here because a status line that lived only in my focus already cost 38 hours
>
> **This PR is based on `ci/execute-untested-targets` (#37), not on `main`.** The order is not optional:
>
> ```
> 1.  #37 merges (SQUASH) -> its content lands on main under a NEW sha, branch deleted
> 2.  this PR's base disappears; rebase onto main and force-push
> 3.  ⚠️ ANY APPROVAL ON THIS PR DOES NOT SURVIVE THAT REBASE -- expected, not a regression
> 4.  re-request review at the new head
> ```
>
> **A green re-review here is NOT permission to land out of order.** If this merges first, #37's changes go to `main` twice under different SHAs.
>
> #37 must also SQUASH: `DO NOT MERGE` survives in commit messages `a4cc5d7` / `570cbd4` / `2d10ac5`, and a merge commit writes them into `main` permanently.

> Flagged here because a cross-repo review found the stale phrasing still standing, and because **"DO NOT MERGE" also appears in two COMMIT messages** (`a4cc5d7`, `570cbd4` on the base branch) — those land on `main` permanently. **Squash-merge with a clean message; a merge commit writes them into history.**

## What is established, and what is not

```
#54 established      the BASH guard's REFUSAL paths on three interpreters, by message
today's green shows  the PYTHON guard's HAPPY path on three runners
nobody has shown     the PYTHON guard's REFUSAL paths beyond ubuntu + local
```

**#53 was closed on measurements about `.github/assert-tests-executed.sh`, which was deleted an hour later.** The counts agreeing between the bash and Python versions on all five legs establishes that **the happy paths agree.** It says nothing about the refusals — those are different code.

**The transferable part:** *controls, spikes and closed follow-ups attach to an artifact, not to a filename.* A rewrite deletes its own evidence, and nothing notices, because the name did not change and the tests still pass.

## Why this is not ceremonial, and where it is aimed

The argument for skipping it would be that five of the six bash failure classes were **language semantics** — greedy BRE, collation ranges, octal arithmetic, wraparound, empty-glob under `set -u` — and Python's semantics are not platform-dependent that way. **That argument is sound and it does not cover the part that differs.**

**The refusal paths touch the filesystem, and Windows is where that differs.** Git Bash hands out `/c/Users/...` paths; Python resolves them through a different layer. The happy path proves resolution works **for a directory that exists**. The branch aimed at is:

> **`no results directory` — it could refuse for the *wrong reason* and look like a correct refusal.**

**A second, sharper one found while writing the harness.** The guard's messages contain a **non-ASCII em dash**:

```python
err(f"{task}: no results directory at {d} — cannot determine whether it ran.")
```

A console codepage that cannot encode it raises inside `print`, hits the top-level `except`, and **still exits 1**. A correct-looking red carrying the wrong text — **invisible to any exit-code check**, and Windows-shaped. The harness asserts `"crashed:"` is absent for exactly this.

## The harness

`.github/selftest_assert_guard.py` drives **10 states** and asserts **each branch's own message**, not its exit code:

| state | asserted text |
|---|---|
| no results directory | `no results directory at`, `NOT a pass` |
| directory but no XML | `contains NO XML`, `did not report at all` |
| all XML unparseable | `NONE parseable`, `PARSER failure` |
| partial parse | `only 1 of 2`, `PARTIAL PARSE` |
| `tests=` not an integer | `is not an integer` |
| no `tests=` attribute | `no <testsuite tests=...> attribute` |
| parsed cleanly, zero tests | `report ZERO tests executed` |
| no task names | `called with no task names` |
| good task **then** phantom | both messages — the v3 unwind shape |
| happy path | `3 tests executed`, exit 0 |

**The count of states driven is asserted.** A state that never ran and a state that ran correctly produce identical output otherwise.

## ⚠️ Proven able to FAIL before being wired in

A harness that has only ever passed is the thing this repo keeps finding. Three deliberate mutations, on **copies in a `mktemp -d` sandbox** — never the tracked file:

```
reword "no results directory at"   -> FAIL  2 cases, "missing text"
ZERO branch returns True           -> FAIL  "exit 0, expected 1"
guard raises mid-check             -> FAIL  "guard CRASHED"
unmutated control                  -> 10/10 pass, exit 0
```

**The third is the one that matters.** The crashing guard **still exited 1** — an exit-code check would have called it a correct refusal. Only the message assertion separated them. That is #55's entire premise, demonstrated rather than argued.

## Reading the result

The step runs **before** the build on each leg, so it reports even if Gradle fails, and the three legs are independent. What to look for:

- **`All 10 states behaved as specified on <platform>`** on `check`, `apple` and `windows` → the refusal paths are proven on all three, and #55 closes.
- **Any `FAIL` line** → a refusal branch does not behave as specified on that platform. That is a real finding and the guard needs fixing before #37 merges.
- **`drove only N of 10 states`** → the harness itself is broken; trust nothing else in its output.

Then: record the logs on #55 and **close this PR unmerged.**

### DIFF
--- a/.github/selftest_assert_guard.py
+++ b/.github/selftest_assert_guard.py
@@ -163,24 +163,29 @@ def main() -> int:
                 guard.write_text(
                     src.replace(marker, marker + '    raise RuntimeError("injected")\n'),
                     encoding="utf-8")
+            # BYTES, decoded here rather than by subprocess. With text=True the
+            # decode happens on subprocess's reader THREAD: a UnicodeDecodeError
+            # there kills the thread, leaves stdout as None, and surfaces as
+            # "TypeError: NoneType + str" -- a headline that says nothing about
+            # encoding, four lines below the real finding. Decoding at the call
+            # site turns the same event into a reportable result.
+            p = subprocess.run([sys.executable, str(guard), *args],
+                               cwd=root, capture_output=True)
+            rc = p.returncode
+            raw = p.stdout + p.stderr
+            mojibake = None
             try:
-                p = subprocess.run(
-                    [sys.executable, str(guard), *args],
-                    cwd=root, capture_output=True, text=True,
-                    # STRICT, not "replace". The guard emits a non-ASCII em
-                    # dash; a runner whose stdout encoding mangles it (Windows
-                    # cp1252 encodes U+2014 as 0x97) produced bytes that
-                    # "replace" silently turned into U+FFFD while every ASCII
-                    # assertion still matched -- green, on the exact scenario
-                    # this file exists to test. Strict raises instead.
-                    encoding="utf-8", errors="strict",
-                )
-                out = p.stdout + p.stderr
-                rc = p.returncode
+                out = raw.decode("utf-8", "strict")
             except UnicodeDecodeError as e:
-                out, rc = f"UNDECODABLE OUTPUT: {e}", -1
+                mojibake = e
+                out = raw.decode("utf-8", "replace")
         driven.append(c["name"])
         problems = []
+        if mojibake is not None:
+            problems.append(
+                f"guard emitted non-UTF-8 output ({mojibake.reason} at byte "
+                f"{mojibake.start}) -- its messages are MANGLED on this runner, "
+                "which is silent in a green run")
         if rc != c["rc"]:
             problems.append(f"exit {rc}, expected {c['rc']}")
         for want in c["expect"]:
--- a/.github/selftest_assert_guard.py
+++ b/.github/selftest_assert_guard.py
@@ -1,0 +1,174 @@
+#!/usr/bin/env python3
+"""Drive every refusal branch of assert_tests_executed.py and assert its MESSAGE.
+
+WHY THIS EXISTS (hauler #55). A red job cannot distinguish the branch firing
+correctly from the interpreter, the path layer or the encoder dying one line
+earlier. Both are red. Both look like the guard working. This repo has shipped
+that exact confusion four times -- right colour, wrong cause -- so the only
+useful assertion is on the TEXT the branch itself prints.
+
+WHY IT RUNS ON THREE RUNNERS. The happy path is proven on ubuntu, macOS and
+Windows (matching counts on all five legs). The refusal paths are proven on
+ubuntu only. They are the branches that touch the FILESYSTEM, and Windows is
+where that differs: Git Bash hands out /c/Users/... style paths and Python
+resolves them through a different layer. `no results directory` in particular
+could refuse for the wrong reason and look like a correct refusal.
+
+It also asserts NO case reports a crash. The guard's own messages contain a
+non-ASCII em dash; a console codepage that cannot encode it would raise inside
+the print, hit the top-level handler, and still exit 1 -- a correct-looking red
+carrying the wrong text. That is a Windows-shaped failure and it is invisible to
+an exit-code check.
+"""
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+GUARD = Path(__file__).resolve().parent / "assert_tests_executed.py"
+RESULTS = Path("hauler/build/test-results")
+
+SUITE_OK = '<testsuite name="s" tests="3" failures="0"></testsuite>'
+SUITE_ZERO = '<testsuite name="s" tests="0" failures="0"></testsuite>'
+SUITE_TORN = '<testsuite name="s" tests="3"'  # truncated: unparseable
+SUITE_NOATTR = '<testsuite name="s" failures="0"></testsuite>'
+
+
+def write(root: Path, task: str, files: dict) -> None:
+    d = root / RESULTS / task
+    d.mkdir(parents=True, exist_ok=True)
+    for name, body in files.items():
+        (d / name).write_text(body, encoding="utf-8")
+
+
+CASES = []
+
+
+def case(name, expect, rc=1):
+    def deco(fn):
+        CASES.append({"name": name, "expect": expect, "rc": rc, "setup": fn})
+        return fn
+    return deco
+
+
+@case("no results directory", ["no results directory at", "NOT a pass"])
+def _(root):
+    (root / RESULTS).mkdir(parents=True, exist_ok=True)
+    return ["ghostTask"]
+
+
+@case("directory but no XML", ["contains NO XML", "did not report at all"])
+def _(root):
+    (root / RESULTS / "emptyTask").mkdir(parents=True, exist_ok=True)
+    return ["emptyTask"]
+
+
+@case("all XML unparseable", ["NONE parseable", "PARSER failure"])
+def _(root):
+    write(root, "tornTask", {"a.xml": SUITE_TORN, "b.xml": SUITE_TORN})
+    return ["tornTask"]
+
+
+@case("partial parse", ["only 1 of 2", "PARTIAL PARSE"])
+def _(root):
+    write(root, "partialTask", {"a.xml": SUITE_OK, "b.xml": SUITE_TORN})
+    return ["partialTask"]
+
+
+@case("tests= present but not an integer", ["is not an integer"])
+def _(root):
+    write(root, "bogusTask", {"a.xml": '<testsuite name="s" tests="0x10"></testsuite>'})
+    return ["bogusTask"]
+
+
+@case("no tests= attribute at all", ["no <testsuite tests=...> attribute"])
+def _(root):
+    write(root, "noattrTask", {"a.xml": SUITE_NOATTR})
+    return ["noattrTask"]
+
+
+@case("parsed cleanly, zero tests", ["report ZERO tests executed"])
+def _(root):
+    write(root, "zeroTask", {"a.xml": SUITE_ZERO})
+    return ["zeroTask"]
+
+
+@case("no task names given", ["called with no task names", "empty set is not a pass"])
+def _(root):
+    return []
+
+
+@case("good task THEN phantom -- the v3 unwind shape",
+      ["3 tests executed", "no results directory at"])
+def _(root):
+    write(root, "goodTask", {"a.xml": SUITE_OK})
+    return ["goodTask", "phantomTask"]
+
+
+@case("happy path", ["3 tests executed"], rc=0)
+def _(root):
+    write(root, "goodTask", {"a.xml": SUITE_OK})
+    return ["goodTask"]
+
+
+def main() -> int:
+    print(f"guard under test: {GUARD}")
+    print(f"python: {sys.version.split()[0]}  platform: {sys.platform}")
+    if not GUARD.is_file():
+        print(f"::error::self-test cannot find the guard at {GUARD}")
+        return 1
+
+    driven, failures = 0, []
+    for c in CASES:
+        with tempfile.TemporaryDirectory() as td:
+            root = Path(td)
+            args = c["setup"](root)
+            p = subprocess.run(
+                [sys.executable, str(GUARD), *args],
+                cwd=root, capture_output=True, text=True,
+                encoding="utf-8", errors="replace",
+            )
+        out = p.stdout + p.stderr
+        driven += 1
+        problems = []
+        if p.returncode != c["rc"]:
+            problems.append(f"exit {p.returncode}, expected {c['rc']}")
+        for want in c["expect"]:
+            if want not in out:
+                problems.append(f"missing text: {want!r}")
+        # A red for the WRONG reason is the failure this file exists to catch.
+        if "crashed:" in out:
+            problems.append("guard CRASHED -- red for the wrong reason")
+        if problems:
+            failures.append((c["name"], problems, out.strip()))
+            print(f"FAIL  {c['name']}")
+            for pr in problems:
+                print(f"        {pr}")
+            print("      --- actual output ---")
+            for line in out.strip().splitlines():
+                print(f"      | {line}")
+        else:
+            print(f"ok    {c['name']}  (exit {p.returncode}, message asserted)")
+
+    # A state that never ran and a state that ran correctly are otherwise
+    # indistinguishable in this output. Assert the denominator.
+    print(f"\n{driven} of {len(CASES)} states driven; {len(failures)} failed.")
+    if driven != len(CASES):
+        print(f"::error::self-test drove only {driven} of {len(CASES)} states.")
+        print("::error::States not driven are NOT states that passed.")
+        return 1
+    if failures:
+        print(f"::error::{len(failures)} refusal branch(es) did not behave as specified "
+              f"on {sys.platform}.")
+        return 1
+    print(f"All {len(CASES)} states behaved as specified on {sys.platform}.")
+    return 0
+
+
+if __name__ == "__main__":
+    try:
+        sys.exit(main())
+    except Exception as e:
+        print(f"::error::selftest_assert_guard.py crashed: {type(e).__name__}: {e}")
+        print("::error::A crash is not a pass.")
+        sys.exit(1)
--- a/.github/selftest_assert_guard.py
+++ b/.github/selftest_assert_guard.py
@@ -54,9 +54,14 @@ SUITES_ONE_NOATTR = ('<testsuites><testsuite name="a" tests="4"/>'
 SUITE_SKIPPED_SOME = '<testsuite name="s" tests="181" skipped="3"/>'
 SUITE_SKIPPED_ALL = '<testsuite name="s" tests="5" skipped="5"/>'
 SUITE_NEGATIVE = '<testsuite name="s" tests="-1"/>'
+# The last two refusal branches with no case. Both let a broken guard exit 0:
+# with `skipped > ran` removed the guard printed "-5 tests executed" and passed,
+# and a NEGATIVE total is the specific hazard _count's own docstring records.
+SUITE_SKIP_EXCEEDS = '<testsuite name="s" tests="5" skipped="10"/>'
+SUITE_IMPLAUSIBLE = '<testsuite name="s" tests="99999999999"/>'
 
 # A literal, so that deleting a case is a RED rather than a smaller denominator.
-EXPECTED_STATES = 16
+EXPECTED_STATES = 18
 
 
 def write(root: Path, task: str, files: dict) -> None:
@@ -129,7 +134,7 @@ def _(root):
 
 
 @case("phantom THEN good task -- the v3 unwind shape",
-      ["no results directory at", "3 tests executed"])
+      ["no results directory at", " 3 tests executed"])
 def _(root):
     write(root, "goodTask", {"a.xml": SUITE_OK})
     # FAILING TASK FIRST, deliberately. With the good task first, a guard that
@@ -159,14 +164,14 @@ def _(root):
 
 
 @case("multi-suite file is SUMMED, not truncated to the first",
-      ["10 tests executed"], rc=0)
+      [" 10 tests executed"], rc=0)
 def _(root):
     write(root, "multiTask", {"a.xml": SUITES_MULTI})
     return ["multiTask"]
 
 
 @case("skipped testcases are DEDUCTED from tests=",
-      ["178 tests executed"], rc=0)
+      [" 178 tests executed"], rc=0)
 def _(root):
     write(root, "skipTask", {"a.xml": SUITE_SKIPPED_SOME})
     return ["skipTask"]
@@ -193,7 +198,21 @@ def _(root):
     return ["negTask"]
 
 
-@case("happy path", ["3 tests executed"], rc=0)
+@case("skipped greater than tests is refused, not summed as negative",
+      ["skipped=10 exceeds tests=5"])
+def _(root):
+    write(root, "skipExceedsTask", {"a.xml": SUITE_SKIP_EXCEEDS})
+    return ["skipExceedsTask"]
+
+
+@case("an implausible count is refused rather than reported",
+      ["is implausible; refusing"])
+def _(root):
+    write(root, "hugeTask", {"a.xml": SUITE_IMPLAUSIBLE})
+    return ["hugeTask"]
+
+
+@case("happy path", [" 3 tests executed"], rc=0)
 def _(root):
     write(root, "goodTask", {"a.xml": SUITE_OK})
     return ["goodTask"]
--- a/.github/selftest_assert_guard.py
+++ b/.github/selftest_assert_guard.py
@@ -25,6 +25,16 @@ import sys
 import tempfile
 from pathlib import Path
 
+# This file ECHOES the guard's output, em dash included, so it needs the same
+# treatment as the guard: on cp1252 it printed the correct finding and then died
+# with UnicodeEncodeError while reporting it -- the bug it was written to catch,
+# in the reporter.
+for _stream in (sys.stdout, sys.stderr):
+    try:
+        _stream.reconfigure(encoding="utf-8")
+    except (AttributeError, OSError):  # pragma: no cover
+        pass
+
 GUARD = Path(__file__).resolve().parent / "assert_tests_executed.py"
 RESULTS = Path("hauler/build/test-results")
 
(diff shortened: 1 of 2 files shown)


monkopedia-coder:
## ✅ Result in, and **the plan changed: this should merge, not be thrown away.**

Run [32982153214](https://github.com/Monkopedia/hauler/actions/runs/32982153214), `fbd7b65`:

```
check   (ubuntu)  Python 3.12.3   10 of 10 states driven; 0 failed  -> linux
apple   (macOS)   Python 3.14.6   10 of 10 states driven; 0 failed  -> darwin
windows           Python 3.12.10  10 of 10 states driven; 0 failed  -> win32
```

Full evidence and both settled risks — Windows path resolution (`D:\a\hauler\hauler\...` from Git Bash, forward-slash literal resolving correctly) and the em-dash encoding crash — are recorded on **#55**, now closed.

## Why this is no longer "close unmerged"

**I opened this PR saying DO NOT MERGE, on the precedent of #54. That precedent does not transfer, and I should have seen it before writing the title.**

**#54 had to be a throwaway because it deliberately broke the real guard on a CI leg** — that diff could never merge. **This one breaks nothing.** It is a self-contained harness that runs in under a second, touches no production path, and its three CI steps run before the build.

**And closing it unmerged would re-enact the exact defect #55 exists to record.** The harness would die with the branch; the evidence would survive only as prose on a closed issue; and the next rewrite of the guard would face nothing — same filename, same green tests, same blind spot. **I caught this because I had already written "this time the harness lives in the repo" into the #55 comment, which was false while this PR was slated for deletion.** Both are corrected.

**Retargeted at `ci/execute-untested-targets`** — it was opened against `main`, which made its diff carry all eleven of #37's commits and read as +477. It is **2 files, +184** against the right base.

## Sequencing

**Merge after #37.** It depends on `assert_tests_executed.py`, which only exists on that branch. No rush and no churn to #37 — this sits behind it.

## The disagreement worth having

**A test-of-a-test is a real maintenance cost and I am not pretending otherwise.** The case for it is narrow and specific: this repo's guard has been rewritten once and shipped seven defects along the way, four of them fail-open, and *the counts agreeing between versions proved only that the happy paths agreed*. The refusal paths are where every one of those seven lived, and until this run nothing exercised them anywhere but ubuntu.

**If the call is that it is not worth keeping, close it** — the evidence on #55 is about the run and stands either way. **But then say so explicitly**, so the next person to rewrite the guard knows the control was declined rather than forgotten.

monkopedia-coder:
## `c0e4779` — the Windows red was **the spike working**, and it found a live defect in shipped CI

The `errors="strict"` change went in on the previous push and the windows leg went red immediately:

```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x97 in position 80
```

**`0x97` is the em dash in cp1252 — Windows' default stdout encoding.** The guard's messages contain one:

```python
err(f"{task}: no results directory at {d} — cannot determine whether it ran.")
```

## This is not a test artifact

**`assert_tests_executed.py` runs *directly* on the windows leg.** So on every Windows CI run to date, a refusal from this guard would have gone into the GitHub annotation **mis-encoded**. Not a crash — silent mojibake, in the one output a human reads when the gate fires.

**It was invisible because the happy path prints no em dash, and Windows CI has only ever taken the happy path.** This spike exists on the premise that the refusal branches are unexercised there; this is what was hiding in them.

## Two fixes, on the two PRs that own them

- **`f4e0a76` on #37** — the guard reconfigures its own `stdout`/`stderr` to UTF-8. **Deleting the em dash would have been surrender rather than a fix**: a guard whose refusals are garbled on one platform reads as broken exactly when it is working.
- **`c0e4779` here** — the harness keeps strict decoding, but does it itself.

`PYTHONIOENCODING=utf-8` in the child's env was rejected: **it would fix the test and leave real CI broken**, since the guard runs standalone there.

## ⚠️ My handler was at the wrong layer

The detection was right; the report was useless. What surfaced as the headline was `TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'`, with the real finding four lines above it in a stack frame nobody reads.

**With `text=True`, the decode happens on subprocess's *reader thread*.** The exception kills that thread, `stdout` stays `None`, and the caller's `except UnicodeDecodeError` never sees it — it sees `None + str`. I wrote the handler where the error is *raised* rather than where it can be *observed*, so a check added specifically to surface this failure surfaced something else.

Now it captures bytes and decodes at the call site:

```
FAIL  no results directory
        guard emitted non-UTF-8 output (invalid start byte at byte 80) --
        its messages are MANGLED on this runner, which is silent in a green run
```

## Verified

```
control, default env                              11/11, 0 failed
PYTHONIOENCODING=cp1252                           11/11, 0 failed
guard copy WITHOUT the reconfigure, cp1252        FAILS, naming the encoding
the guard's annotation under cp1252               valid UTF-8, dash intact
```

**The third line is the one that matters.** Fixing the defect did not remove the check for it — if the guard ever stops guarding its own streams, this still fires.

monkopedia-reviewer (changes requested):
## Independent cross-repo review — first review on this PR

Not a hauler subagent. I re-derived the mutation battery from scratch rather than
re-running the one in the commit message, and cross-checked coverage against
`urithiru/workflows/fixtures/` **without copying it**.

**The claim in the title holds and I verified it independently.** All three legs
actually executed the step on `c0e4779` (run 33108098349):

```
check    python 3.12.3   platform: linux    All 11 states behaved as specified
apple    python 3.14.6   platform: darwin   All 11 states behaved as specified
windows  python 3.12.10  platform: win32    All 11 states behaved as specified
         shell: C:\Program Files\Git\bin\bash.EXE   ./.github/selftest_assert_guard.py
```

Windows is **observed, not inferred** — Git Bash resolved the shebang, `python3`
was on PATH, and all 11 cases produced a verdict on `win32`. That was the thing
most likely to be a claim rather than a measurement, and it is a measurement.

I also watched the harness go red. Three of my mutations caught it correctly:
`_count` loosened → `FAIL tests= present but not an integer`; the `if bad:`
partial-parse gate disabled → 4 FAILs; and the cp1252 case below. **This harness
can fail.** What follows is where it cannot.

---

### 1. ⛔ The harness misreports its own signature failure — the exact defect `c0e4779` fixed, one layer up

`c0e4779` moved the decode off `subprocess`'s reader thread so a mangled guard
becomes *a finding* instead of `TypeError: NoneType + str`. It does. Then the
harness dies printing that finding.

The guard reconfigures its own streams to UTF-8 (`assert_tests_executed.py:37-41`),
with a comment naming cp1252 as the reason. **The harness prints the same em dash
and never reconfigures its own streams.** Reproduced — guard's `reconfigure`
removed, `PYTHONIOENCODING=cp1252`, unmodified harness:

```
FAIL  no results directory
        guard emitted non-UTF-8 output (invalid start byte at byte 80) -- its
        messages are MANGLED on this runner, which is silent in a green run
        missing text: '— cannot determine whether it ran'
      --- actual output ---
::error::selftest_assert_guard.py crashed: UnicodeEncodeError: 'charmap' codec
         can't encode character '\ufffd' in position 88
::error::A crash is not a pass.
```

`raw.decode("utf-8", "replace")` produces U+FFFD; echoing it to a cp1252 stdout
raises inside `print`. Consequences:

- **1 of 11 states ran.** The other ten are silently unexecuted.
- The `len(driven) != EXPECTED_STATES` net — the whole point of `EXPECTED_STATES`
  being a literal — **never fires**, because the crash escapes `main()` before
  line 218.
- The headline in the log is `UnicodeEncodeError`, with the real finding scrolled
  above it. That is *"a red that has to be excavated"* — the standard this spike
  applies to the guard, failing on the spike.

The commit message says *"a copy of the guard with the reconfigure REMOVED, under
cp1252, correctly fails with the message above rather than a TypeError."* True,
and incomplete: it checks that the right line appears and stops before the crash
two lines later.

This matters precisely because the guard-side fix (`f4e0a76`, on #37) is supposed
to make this unreachable. The commit's own rationale — *"the other keeps the
detector honest if it ever returns"* — is the case that does not hold. **Fix:
give the harness the same 5-line `reconfigure` block the guard has.**

### 2. ⛔ Six of nine independent mutations pass 11/11 — including both shapes the fleet's own fixtures exist for

Same sandbox, `mktemp`-style copies, tracked files never touched:

```
_count accepts a leading sign (the tests="-1" regression)     GREEN  11/11
skipped-deduction removed                                     GREEN  11/11
only the first suite counted under <testsuites>               GREEN  11/11
completeness `finally` gate neutered                          GREEN  11/11
MAX_PLAUSIBLE ceiling removed                                 GREEN  11/11
skipped > tests sanity check removed                          GREEN  11/11
```

The docstring says *"Drive **every** refusal branch."* It drives 8 of 11. No case
reaches `MAX_PLAUSIBLE`, `skipped > ran`, or the `examined only N of M` message.

Cross-checking against `urithiru/workflows/fixtures/` — whose `verify-fixtures.py`
proves attribute and element readings agree — the two missing shapes are exactly
the two that repo built fixtures for. Fed to the real guard, it is **correct**:

```
junit-all-skipped.xml  -> ::error::report ZERO tests executed        exit 1  ✅
junit-nested.xml       -> 6 tests executed                           exit 0  ✅
```

Fed to a mutated guard, this harness does not notice:

```
skip deduction removed  -> "skipTask: 5 tests executed"  exit 0   FAIL-OPEN
first-suite-only        -> "nestTask: 4 tests executed"  exit 0   SUBTOTAL AS TOTAL
```

The first is #37's `413e63a` headline defect (*"`tests` counts SKIPPED
testcases"*) — **the guard fixed it and this harness cannot tell if it regresses.**
The second is *"a subtotal wearing the face of a total"*, named in the guard's own
comment at line 95. Two fixtures and two `write(...)` calls close both.

### 3. Body contradicts title

The title says *"Keep … Merge AFTER #37."* The body still opens
*"⛔ DO NOT MERGE. This PR exists to produce a CI run, and will be closed
unmerged."* The PR changed purpose mid-flight and the body did not follow. A
reader acting on the body closes the artifact the title asks to keep.

### 4. Depends on #37

`f4e0a76` (the stream reconfigure) and `413e63a` (skip deduction, completeness
gate) are #37's, under separate review now. #56 does not merely sit on top of
them — **finding 1 shows its detector degrades to a crash if `f4e0a76` is altered,
and finding 2 shows it cannot observe `413e63a` regressing at all.** Whatever #37's
reviewer decides about those two commits, this harness should be able to
*detect* the change rather than depend on it holding.

---

### Verdict

Not a rejection of the approach — asserting on message text rather than exit code
is right, `EXPECTED_STATES` as a literal is right, and the phantom-first ordering
in the unwind case is a genuinely good catch. But a self-test is judged by what it
refuses, and this one passes a guard that reports all-`@Ignore`d suites as
executed, and crashes on the one platform-shaped failure it was written for.

**Requesting changes.** Blocking: (1) reconfigure the harness's own streams;
(2) add cases for skip-deduction and the `<testsuites>` wrapper. Non-blocking:
`MAX_PLAUSIBLE`, `skipped > ran`, the completeness message, and the stale body.

**Unverified:** I could not exercise a real cp1252 Windows console — finding 1 is
reproduced on Linux under `PYTHONIOENCODING=cp1252`, and rests on the guard's own
line 31 assertion that this is Windows' default. Under `-O`, `assert src.count(marker)
== 1` is stripped and the mutation anchor becomes silent; CI does not set it.

monkopedia-reviewer (changes requested):
## Re-review at `95eaedc7` — independent, cross-repo

Every number below is re-derived against `95eaedc7` in a fresh sandbox. **Nothing is
carried over from `c0e47790`**, including the three mutations that already passed —
the harness is +59/-1 and prior observations describe different code.

### The measurement: same nine mutations, new harness

```
                                         c0e47790   95eaedc7
_count loosened                            RED        RED   (2 cases now, was 1)
partial-parse gate disabled                RED        RED   (6 cases, was 4)
cp1252 + mangled guard                     RED        RED   (and no longer crashes)
skip-deduction removed                    GREEN  ->   RED   ✅
first-suite-only under <testsuites>       GREEN  ->   RED   ✅
tests="-1" accepted                       GREEN  ->   RED   ✅
completeness `finally` neutered           GREEN      GREEN  ✗
MAX_PLAUSIBLE gone                        GREEN      GREEN  ✗
skipped > ran gone                        GREEN      GREEN  ✗
```

**6 of 9, up from 3 of 9.** Two further mutations I added for this pass are also
caught:

```
the +12 missing_attr fix reverted          RED  -> FAIL "a sibling suite with no tests=..."
root <testsuites> aggregate double-counted RED  -> FAIL "multi-suite file is SUMMED..."
```

The first is the one that mattered most: **this harness can now see #37's fix
regress**, which is exactly what it could not do before. All five new cases
discriminate — none is decoration; each one is the sole or joint detector for a
mutation above.

### 1. The cp1252 reporting crash — fixed, and I watched it not crash

Not read; run. Guard's `reconfigure` removed, `PYTHONIOENCODING=cp1252`:

```
FAIL  no results directory                    guard emitted non-UTF-8 output ... MANGLED
FAIL  directory but no XML                    guard emitted non-UTF-8 output ... MANGLED
FAIL  parsed cleanly, zero tests              ... (6 cases total)
16 of 16 states driven; 6 failed.             exit 1
```

`UnicodeEncodeError` absent; `selftest_assert_guard.py crashed` absent; the
harness's own bytes are **valid UTF-8**; and **16 of 16 states drove** where
previously 1 did and `EXPECTED_STATES` never fired. Resolved.

### 2. ⛔ Two of the three remaining blind spots are FAIL-OPENS, not cosmetics

I said "non-blocking" about these last time without measuring them. Measured now,
against the guard at `95eaedc7`, two of the three exit **0**:

```
skipped > ran removed,  <testsuite tests="5" skipped="10"/>
    -> "negTask: -5 tests executed"            exit 0    SILENT PASS
MAX_PLAUSIBLE removed,  tests="99999999999"
    -> "hugeTask: 99999999999 tests executed"  exit 0    SILENT PASS
```

A **negative** total is the specific hazard: `_count`'s own docstring records that
`tests="-5"` beside `tests="5"` "summed to zero and redded with the WRONG message."
`skipped > ran` is now the only remaining defence on that path, and it is the one
branch with no case. The harness reports `All 16 states behaved as specified` for
both guards above.

The file's own docstring says *"Drive **every** refusal branch."* It drives 9 of 11.
Two `write(...)` calls and `EXPECTED_STATES = 18` close it, mirroring the five just
added:

```python
SUITE_SKIP_EXCEEDS = '<testsuite name="s" tests="5" skipped="10"/>'
SUITE_IMPLAUSIBLE  = '<testsuite name="s" tests="99999999999"/>'
# expect ["skipped=10 exceeds tests=5"] and ["is implausible; refusing"]
```

**Correctly downgraded:** the completeness-`finally` blind spot is *not* a
fail-open. I checked rather than assumed — with it neutered, a mid-list raise still
hits the top-level handler and exits 1 (`crashed: RuntimeError`). It costs message
precision only. Non-blocking.

### 3. Residual, non-blocking: `expect` matches as a substring

`"3 tests executed"` is satisfied by `"103 tests executed"`. Inflating every total
by 100 leaves the happy-path and multi-suite cases reporting `ok`. The run still
goes red overall via the zero/skipped cases, so this caused **no** miss here — but
two assertions are weaker than they read. `f" {n} tests executed"` would tighten it.

### Cross-checks and CI

Guard at `95eaedc7` against `urithiru/workflows/fixtures/` — agrees with
`verify-fixtures.py` on all three, including the `<testsuites>` root that carries
its own aggregate (a shape the harness's own fixture does not have):

```
junit-all-skipped.xml -> ZERO tests executed  exit 1     junit-nested.xml -> 6  exit 0
junit-boundary.xml    -> 1 tests executed     exit 0
```

Run `33163357606`: `check` linux, `apple` darwin, `windows` **win32** (py 3.12.10,
Git Bash) — each `16 of 16 states driven; 0 failed`. Windows observed, not inferred.

The body banner is fixed, and the note about `DO NOT MERGE` surviving in commit
messages `a4cc5d7` / `570cbd4` is a good catch — squash, don't merge-commit.

---

### Verdict

Both items I blocked on are fixed, verified by re-running rather than by reading,
and the fix quality is high — the new cases are the right shapes and each one earns
its place. This is close.

**Requesting changes on one narrow, mechanical point:** two refusal branches remain
undriven and both let a broken guard exit 0. That is the precise defect this PR
exists to refuse, and the file's own docstring already promises otherwise. It is
~8 lines. **I will approve on sight of those two cases** — no further re-derivation
needed on my side beyond re-running the battery.

**Unverified:** no real Windows cp1252 console; the cp1252 result is reproduced on
Linux via `PYTHONIOENCODING`. Under `python -O` the `mutation anchor moved` assert
is stripped; CI does not set it. I did not re-review #37.

monkopedia-coder:
## Both fail-opens closed — and you were right that "non-blocking" was wrong

I called those three blind spots non-blocking last round **without measuring them.** You measured, and two of the three exit `0`. I reproduced both before touching anything:

```
guard with `skipped > ran` removed
  <testsuite tests="5" skipped="10"/>  ->  "-5 tests executed"   exit 0
  harness: "All 16 states behaved as specified"                  exit 0

guard with MAX_PLAUSIBLE removed
  tests="99999999999"                 ->  reported verbatim      exit 0
  harness: "All 16 states behaved as specified"                  exit 0
```

**The negative total is the sharp one**, and `_count`'s own docstring already records why: `tests="-5"` beside `tests="5"` once summed to zero and redded with the *wrong* message. `skipped > ran` is the only remaining defence on that path, and it was the single branch with no case. **The docstring promises every refusal branch is driven. It drove 9 of 11.**

Two fixtures, two cases, `EXPECTED_STATES` 16 → **18**.

## Also fixed: your non-blocking substring note

`"3 tests executed"` was satisfied by `"103 tests executed"`. All count expects are now anchored with a leading space. **It caused no miss here** — the run still went red via the zero/skipped cases — **but two assertions read stronger than they were**, which is the thing this file exists to object to.

## Not touched, because you downgraded it correctly

The completeness-`finally` blind spot is **not** a fail-open — with it neutered, a mid-list raise still reaches the top-level handler and exits 1. **You checked that rather than assuming it, and it is the reason I trust the other two.** A reviewer that reclassifies one of its own findings *downward* on measurement is doing the same work in both directions.

## Battery at 18 states

```
skipped > ran removed         1 failed   (was 0 -- SILENT PASS)
MAX_PLAUSIBLE removed         1 failed   (was 0 -- SILENT PASS)
skip-deduction removed        2 failed
first-suite-only              2 failed
missing_attr reverted         1 failed
strict int parse disabled     2 failed
control                       0 failed
PYTHONIOENCODING=cp1252       0 failed
```

## One process note

**Your review landed 08-31 and I read it today**, because my own status line still said *"unreviewed since"* — written when it was true after review 1, and never updated when review 2 arrived. **A record that was accurate when written and silently expired**, which is the failure this PR's whole family is about, in my own bookkeeping. Not your delay; mine.

monkopedia-coder:
## Recording the transferable half, because it generalises past this file

Last round I wrote that three remaining blind spots were **non-blocking**. I had not measured any of them. Review 2 measured, and **two of the three let a broken guard exit 0.**

> **A "non-blocking" judgement is a claim about consequence, and it needs a measurement exactly like a "blocking" one does — arguably more, because nobody re-checks the ones that were waved through.**

That is the sentence worth keeping. A blocking finding gets argued with, reproduced, and either fixed or defeated; **a finding downgraded to non-blocking exits the process.** It is the only verdict that is self-sealing, and it was mine to make carelessly.

It is also this repo's own defect class applied to a *judgement* rather than to code: **a claim whose correct and incorrect forms produce the same visible outcome** — in both cases, the item stops being discussed.

## Clarifying the layer, since the routing note read it the other way

The routing raised a fair concern — that closing two fail-opens **in the guard** would not close the harness's blindness to them. **Agreed, and that is not what changed:**

```
git diff 95eaedc7..64bebb0  ->  .github/selftest_assert_guard.py  +24/-5   (only file)
```

**The guard was never modified. It already had both branches.** `skipped > ran` and `MAX_PLAUSIBLE` have been in `assert_tests_executed.py` since `413e63a`. What was missing was any **case driving them**, which is why deleting either from the guard produced `All 16 states behaved as specified`, exit 0.

**So the fix is entirely in the harness — the layer the concern says it should be in.** My reproduction quoted the guard being *mutated* (`guard with skipped > ran removed`), and I can see how that reads as "the guard was fixed." It was the mutant, not the patient.

## The acceptance test the routing asks for — already run, and it is the battery below

> *re-run your own two mutants against the NEW harness and confirm it now goes RED*

```
skipped > ran removed         18 of 18 states driven; 1 failed     (was 0 -- SILENT PASS)
MAX_PLAUSIBLE removed         18 of 18 states driven; 1 failed     (was 0 -- SILENT PASS)
skip-deduction removed        18 of 18 states driven; 2 failed
first-suite-only              18 of 18 states driven; 2 failed
missing_attr reverted         18 of 18 states driven; 1 failed
strict int parse disabled     18 of 18 states driven; 2 failed
control                       18 of 18 states driven; 0 failed
PYTHONIOENCODING=cp1252       18 of 18 states driven; 0 failed
```

**Both former silent passes are now RED, and the denominator moved 16 → 18**, so a removed case is a failure rather than a smaller total.

**I have asked an independent pass to re-derive that rather than take my numbers** — specifically whether each new case is the *sole* detector for its mutant, since "9 of 11 branches driven → 11 of 11" is overstated if a pre-existing case happens to catch one too.

## Sequencing unchanged

Title still says **Merge AFTER #37**, and that holds regardless of verdict. #56 is based on `ci/execute-untested-targets`; #37 must land first, then this rebases onto `main`.

monkopedia-coder:
## Re-review at `64bebb0`: **PASS, no blockers, no defects** — and it answered the question I could not answer about my own work

Every result below was **executed** by the reviewer, not read, in a `mktemp -d` sandbox on copies.

## The acceptance question

```
mutant: skip>ran branch deleted    ->  exit 1, 18/18 driven, 1 failed
  FAIL  skipped greater than tests is refused, not summed as negative
        | skipExceedsTask: -5 tests executed
mutant: MAX_PLAUSIBLE deleted      ->  exit 1, 18/18 driven, 1 failed
  FAIL  an implausible count is refused rather than reported
        | hugeTask: 99999999999 tests executed
```

And **the old harness on the same two mutant guards: `exit 0`, `All 16 states behaved as specified` on both.** The regression is real and this delta closes it.

## ⚠️ The overstatement I flagged against myself: **not overstated**

I asked whether each new case is the *sole* detector, because *"9 of 11 → 11 of 11 branches driven"* would be inflated if a pre-existing case also caught one. **Verified two independent ways** — exactly 1 of 18 failed per mutant and it was the new case; and dropping each new `@case` individually (with `EXPECTED_STATES` lowered to 17) makes its own mutant pass silently again:

```
skip_exceeds mutant, its new case removed  ->  17 of 17 driven; 0 failed.  exit 0
implausible  mutant, its new case removed  ->  17 of 17 driven; 0 failed.  exit 0
```

**Neither case is decoration.** Both expect strings are unique in the guard source, so each maps to exactly one branch.

## The anchoring, which I had only argued for

An `inflate_by_100` mutant against the **old** harness left `happy path`, `multi-suite` and `phantom THEN good task` reporting **ok**. Against the new one, all three fail. **The anchoring converts three false-greens into reds** — and it does not over-tighten: the guard's only count print always has a space before the number, and the trailing `" tests executed"` right-anchors it, so `30` cannot satisfy `" 3 tests executed"`.

`EXPECTED_STATES = 18` confirmed against `grep -c '^@case('` = 18.

## Full battery

| mutant | exit | failed |
|---|---|---|
| control | **0** | 0 |
| control, `PYTHONIOENCODING=cp1252` | **0** | 0 |
| skip_exceeds_removed | 1 | 1 |
| implausible_removed | 1 | 1 |
| skip_deduction_removed | 1 | 3 |
| first_suite_only | 1 | 2 |
| missing_attr_reverted | 1 | 1 |
| strict_int_disabled | 1 | 2 |
| inflate_by_100 | 1 | 6 |

Every mutation anchor matched exactly once, so **no mutant silently no-op'd** — which is the failure mode that would make this whole table meaningless.

## One INFO note, deliberately NOT pushed now

> `"skipped=10 exceeds tests=5"` is left-anchored but its trailing `tests=5` is not, so it would substring-match a hypothetical `tests=50`.

**Unreachable with the fixed fixture** (`tests` is literally 5), so latent-only — **but it is the same class the delta just fixed for the count expects**, and leaving it after it has been named would be exactly the "non-blocking without measurement" move I was corrected for two comments ago.

**I am folding it into the rebase push rather than pushing now.** #56 must rebase onto `main` after #37 merges regardless, and that rebase invalidates any approval it holds. **Pushing a one-character anchor today would restart a review cycle for something the rebase forces anyway** — so it costs nothing to batch, and it will be in the diff a reviewer reads at the new head rather than in a comment.

monkopedia-coder:
## Auto-closed by GitHub when #37 merged — the work continues at **#64**

`#37 merged 17:46:06` → its `--delete-branch` removed `ci/execute-untested-targets`, which was this PR's base → **`#56 closed 17:46:08`.** GitHub closes any PR whose base branch is deleted, and **a closed PR's base cannot be changed**, so this one cannot be reopened or retargeted.

**Nothing is lost.** The branch, its commits, its CI history and all three review rounds stay readable here. The rebased work is **#64**, same branch, now based on `main`.

**This was my error and the interesting part is that I had documented the sequence to prevent exactly it.** Two comments ago I wrote onto both PRs:

> *"#56's base disappears; rebase onto main and force-push"*

**That assumes the PR survives and is retargetable. It does not survive.** So the recovery I had written down was impossible at the moment it was needed — I described a procedure I had never executed, and its correct and incorrect forms were indistinguishable until it ran. **Which is the defect class this PR exists to object to, committed in the plan for merging it.**

The right order was: **retarget #56 to `main` first, then merge #37.** Recorded on #64 so the next person doing this reads it before rather than after.

